PR #119, opened 2021-09-17
Adds testing using github actions

### LINKED ISSUE #118: set up Github actions
We need to add automated testing pipelines again since moving from bitbucket.

- [x] add github actions
- [x] remove btibucket pipelines

laurencejackson:
I've added three Github actions to the repo. 

1. Run pytest - currently have lint testing commented out since we didn't do it before and the code isn't ready for it yet! I think this is a test we should definitely look to having in the future. 
2. Deployment - this action builds, tags, and pushes an image to dockerhub when a tagged release event is detected by github
3. Added dependabot - GitHub bot that should detect out of date actions


hshuaib90 (approved):
LGTM 🥳

PR #122, opened 2021-10-12
Add GitHub actions

laurencejackson:
Modifies github actions to include automatic reporting of coverage and linting to PRs

The checks currently fail, but that's because the linting action has found an undefined variable that correctly results in the failed result.


laurencejackson:
This is ready for review now, sorry about how many coverage reports the bot has made with comments - this was part of the debugging. You can take the most recent one as what to expect.

It looks like currently, the branch coverage function isn't compatible with the PR annotation action. I think it just takes the bottom row of the 3rd column of the coverage report - so if we have more than 3 columns (like when using branch coverage) the 3rd value is not coverage%.

hshuaib90 (approved):

PR #123, opened 2021-10-12
Origin/feature/hzn 80   function to set log level from CLI

### LINKED ISSUE #80: [HZN-80] Add logging to screen
Transferred by @LaurenceJackson on Aug 13, 2020, 11:31

:::Issue recreated from Jira issue [HZN-80](https://gsttmri.atlassian.net/browse/HZN-80):::

Add python logger to print to screen log messages.

Add CLI for setting verbosity (e.g. info, debug, warn, error)

Lucrezia-Cester:
added logging which can now replace printing. Use log("message") instead of print("message"). To set log level from CLI check log_levels_fun -h.


laurencejackson:
Thanks for this Lu! This is a great start but could we approach this from another direction? I think what we want to do is to create a logger class that is instantiated and can then be used from any python file where it's required (i.e. the hazenlib functions). For an example of how to create the Logger class take a look at the dicomserver logger on bitbucket, although this has many features we're unlikely to need in this case so you might be better off starting [here](https://docs.python.org/3/howto/logging.html#logging-basic-tutorial). The "Logging from multiple modules" bit should be helpful. 

Could you also add some logging statements to the hazenlib functions? I think we want:

- INFO level logs each time a major function is called (e.g. starting tasks like snr, slice_position etc)
- DEBUG level logs when lower-level functions are called, e.g. functions in hazenlib (including in hazenlib/__init__.py)
- replace print statements in hazenlib with logging statements (probably most of these will be DEBUG level logs but you can judge whether you feel any should be INFO level) 

To meet the requirement of being able to change the logging level from the CLI the best way to go about this would be to add a new parameter to the hazen CLI in the same way as the report flag- e.g.

`hazen <task> <folder> [--measured_slice_width=<mm>] [--report] [--debug level]`

Lucrezia-Cester:
Hi Laurence, I tried following your instructions. Now one can pass a logging level as an optional argument from the Hazen CLI. If nothing is passed a default logging level is set. I only replaced some previous print statements with log statements. I will add more logging statements as I keep working on Hazen.

hshuaib90:
There are currently merge conflicts on this branch. @laurencejackson  are you able to work with @Lucrezia-Cester to get this fixed?

hshuaib90 (changes requested):
The True_COL_top.png and True_COL_right.png files need to be uncommitted.

Lucrezia-Cester:
Hi @laurencejackson , the code is now ready to be reviewed

laurencejackson:
Hi @Lucrezia-Cester, could you investigate why the tests are failing? It looks like you're trying to import a package called 'logger' but there is no package with that name. Could you mean to import 'logging'? or are you trying to import a logger object from another file? in which case you will want to do something like "from x import logger".

let me know if i can help.

Lucrezia-Cester:
Hi @laurencejackson , basically I created a logger.py file with a logger function (it's your MLOP logger) and that is what I was trying to import.

hazenlib/slice_position.py
```diff
@@ -4,6 +4,7 @@
 http://scikit-image.org/docs/0.11.x/auto_examples/plot_local_otsu.html
 
 """
+from logger import logger
```
laurencejackson: For this to work it will need to know where the logger package is. In our case the logger object we want is in hazenlib/logger.py so this will need to read:

`from hazenlib.logger import logger`

hazenlib/slice_width.py
```diff
@@ -7,6 +7,7 @@
 import sys
 import traceback
 from copy import copy
+from logger import logger
```
laurencejackson: as above

hazenlib/spatial_resolution.py
```diff
@@ -12,6 +12,9 @@
 import copy
 import sys
 import traceback
+from logger import logger
```
laurencejackson: as above

hazenlib/slice_position.py
Lucrezia-Cester: ah ok thanks!! will change it to that then!

laurencejackson (changes requested):
Nearly there I think! Could you remove the .png files you've accidentally committed and make the changes and replace 

`from logger import logger`

with

`from hazenlib.logger import logger`

laurencejackson:
@Lucrezia-Cester could you add the `colorlog` pacakge to the hazen requirements.txt? IT looks like this is whats causing the tests to fail.

hshuaib90:
Is this now resolved?

hazenlib/spatial_resolution.py
```diff
@@ -405,7 +408,7 @@ def get_esf(edge_arr, y):
     return u, esf
 
 
-def calculate_mtf_for_edge(dicom, edge, report_path=False):
+def calculate_mtf_for_edge(dicom, edge, report_path=True):
```
laurencejackson: Is there a reason that this is changed now? I think we want to keep the report_path kwarg default as False.

laurencejackson (changes requested):
Looks good, just one _tiny_ change suggested. It looked like you changed a kwarg - probably for local debugging because the CLI wasn't working before. This issue was fixed in a separate PR so we can put this back to how it was.

hazenlib/spatial_resolution.py
Lucrezia-Cester: hey yes sorry I did it for debugging. So, should I put this back to how it was?

hshuaib90: yep 👍🏽

hshuaib90 (approved):

PR #124, opened 2021-10-14
Feature/hzn 73 sp res mtf plot should only plot half values

### LINKED ISSUE #73: [HZN-73] SpRes-MTF plot should only plot half values

:::Issue recreated from Jira issue [HZN-73](https://gsttmri.atlassian.net/browse/HZN-73):::

Imported from https://bitbucket.org/gsttmri/hazen/issues/74/spres-mtf-plot-should-only-plot-half  Consider only plotting half of the MTF \(L465\)--the second half is a Nyquist aliasing \(I think!\) from the FFT.

Lucrezia-Cester:
Original Issue:
 SpRes-MTF plot should only plot half values #73

  Consider only plotting half of the MTF (L465)--the second half is a Nyquist aliasing (I think!) from the FFT.

Upon inspection, the plot of MTF contained both positive and negative frequencies, because that is what the FFT outputs since the basic functions of the fft are complex exponentials. 
It makes sense to only plot the positive frequencies, so I changed how the MTF was plotted.

### DIFF
--- a/hazenlib/spatial_resolution.py
+++ b/hazenlib/spatial_resolution.py
@@ -424,6 +424,10 @@ def calculate_mtf_for_edge(dicom, edge, report_path=False):
     x, y = get_edge_profile_coords(angle, intercept, spacing)
     u, esf = get_esf(edge_arr, y)
     lsf = maivis_deriv(u, esf)
+    lsf = np.array(lsf)
+    n=lsf.size
+    freqs= fftfreq(n)
+    mask = freqs > 0
     mtf = abs(np.fft.fft(lsf))
     norm_mtf = mtf / mtf[0]
     mtf_50 = min([i for i in range(len(norm_mtf) - 1) if norm_mtf[i] >= 0.5 >= norm_mtf[i + 1]])
@@ -461,7 +465,7 @@ def calculate_mtf_for_edge(dicom, edge, report_path=False):
         axes[9].set_title('line spread function')
         axes[9].plot(lsf)
         axes[10].set_title('normalised MTF')
-        axes[10].plot(norm_mtf[60:120])
+        axes[10].plot(freqs[mask],norm_mtf[mask])
         fig.savefig(f'{report_path}_{pe}_{edge}.png')
 
     return res
--- a/hazenlib/spatial_resolution.py
+++ b/hazenlib/spatial_resolution.py
@@ -427,12 +427,12 @@ def calculate_mtf_for_edge(dicom, edge, report_path=False):
     lsf = maivis_deriv(u, esf)
     lsf = np.array(lsf)
     n=lsf.size
-    mask = freqs >= 0
     mtf = abs(np.fft.fft(lsf))
     norm_mtf = mtf / mtf[0]
     mtf_50 = min([i for i in range(len(norm_mtf) - 1) if norm_mtf[i] >= 0.5 >= norm_mtf[i + 1]])
     profile_length = max(y.flatten()) - min(y.flatten())
     freqs= fftfreq(n, profile_length/n)
+    mask = freqs >= 0
     mtf_frequency = 10.0 * mtf_50 / profile_length
     res = 10 / (2 * mtf_frequency)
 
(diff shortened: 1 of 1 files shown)


hazenlib/spatial_resolution.py
```diff
@@ -461,7 +461,7 @@ def calculate_mtf_for_edge(dicom, edge, report_path=False):
         axes[9].set_title('line spread function')
         axes[9].plot(lsf)
         axes[10].set_title('normalised MTF')
-        axes[10].plot(norm_mtf)
+        axes[10].plot(norm_mtf[60:120])
```
laurencejackson: I think the modulation transfer function (MTF) we want to see is the first half of this array! Could you also make this more generalised? so rather than being `[0:60]` it could be something like `[0:(0.5*len(norm_mtf))]`.

laurencejackson (changes requested):
Thanks Lu, requested some changes - let me know if any issues.

laurencejackson (approved):
Looks great!

hshuaib90:
@pcw24601 since you raised this issue initially, are you ok to review Lucrezia's PR? and then we can merge

hazenlib/spatial_resolution.py
```diff
@@ -424,6 +425,10 @@ def calculate_mtf_for_edge(dicom, edge, report_path=False):
     x, y = get_edge_profile_coords(angle, intercept, spacing)
     u, esf = get_esf(edge_arr, y)
     lsf = maivis_deriv(u, esf)
+    lsf = np.array(lsf)
+    n=lsf.size
+    freqs= fftfreq(n)
+    mask = freqs > 0
```
pcw24601: Can you make this `freqs >= 0 ` to include the first FFT term?

hazenlib/spatial_resolution.py
```diff
@@ -424,6 +425,10 @@ def calculate_mtf_for_edge(dicom, edge, report_path=False):
     x, y = get_edge_profile_coords(angle, intercept, spacing)
     u, esf = get_esf(edge_arr, y)
     lsf = maivis_deriv(u, esf)
+    lsf = np.array(lsf)
+    n=lsf.size
+    freqs= fftfreq(n)
```
pcw24601: This gives `freqs` in the units [1/pixels]. Try `freqs= fftfreq(n, d=profile_length/n)` to get in units [1/mm], which should fix the graph axis (line 469). You'll have to put this after the definition of profile_length in line 435.

This seems to give better units on the x-axis for the 256 and 512 images, and the MTF=0.5 aligns better with what I'd expect, 0.5 lp/mm for the 256 image (i.e. 1 / [2*pixel size 0.98mm]) and 1 lp/mm (1/[2*0.49mm]). However, I'm not really experienced with MTF and happy to be proved wrong!
256:
image

512:
image

hazenlib/spatial_resolution.py
```diff
@@ -461,7 +466,7 @@ def calculate_mtf_for_edge(dicom, edge, report_path=False):
         axes[9].set_title('line spread function')
         axes[9].plot(lsf)
         axes[10].set_title('normalised MTF')
-        axes[10].plot(norm_mtf)
+        axes[10].plot(freqs[mask],norm_mtf[mask])
```
pcw24601, line 469: [Update--see comment on line 430 for possible resolution]

I can't figure out the units on this axis. Are they pixels^(-1)? The MTF graph for a 512 pixel image (tests\data\resolution\RESOLUTION\ax_512_resolution.dcm) and the 256 pixel image (tests\data\resolution\eastkent\256_sag.IMA) are very similar.
512:
image
256:
image

The actual spatial resolution of the 512 image is around 1 line pair /mm and for the 256 image it's about 2 line pairs /mm as can be seen on the comb inserts.

pcw24601 (changes requested):
This is great! I've made a couple of comments and welcome your thoughts.

pcw24601 (approved):
This looks great.

PR #127, opened 2021-10-25
Fix unit tests so failed tests trigger a fail

laurencejackson:
Main issue is caused by piping the output of the pytest function to the pytest-coverage.txt file - this pipeing operation completes successfully whether pytest passes or fails so the exist code will always be 1. By first setting `pipefail -o set` then we can reverse the direction of the exit code evaluation so exit codes are a result of the leftmost function in a command.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -60,7 +60,8 @@ jobs:
 
     - name: Test with pytest
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/python-app.yml
```diff
@@ -4,9 +4,6 @@
 name: Tests
 
 on:
-  push:
-    branches:
-      - '*'
```
laurencejackson, line 9: if we only test on PRs then we prevent triggering this action twice on each push to a branch which has an open PR. This is what causes the duplicate code coverage comments in the PR too.

hshuaib90 (approved):

PR #132, opened 2021-11-02
hazen to ignore non-DICOM files

### LINKED ISSUE #106: [HZN-107] Ignore non-DICOM files in working directory
Transferred by @LaurenceJackson on Jun 17, 2021, 15:35

:::Issue recreated from Jira issue [HZN-107](https://gsttmri.atlassian.net/browse/HZN-107):::

Hazen should ignore non-DICOM files in a directory rather than crash.

Lucrezia-Cester:
hazen to ignore non-DICOM files  instead of crashing

Not sure why it is not letting me link to the issue tho?  issue #106


laurencejackson:
@Lucrezia-Cester is this ready for review?

Lucrezia-Cester:
@laurencejackson yes!

hazenlib/__init__.py
```diff
@@ -298,14 +298,35 @@ def get_field_of_view(dcm: pydicom.Dataset):
 
     return fov
 
+  
+def is_dicom_file(filename):
```
hshuaib90: Can we have a test function that shows this works in an associate test file?

Lucrezia-Cester: @hshuaib90 added, let me know if it is ok!

laurencejackson (changes requested):
Thanks Lu, all looks good to me except the `is_dicom_file` function is now duplicated in hazenlib.__init__ and in hazenlib.tools. 

1. Could you delete the version of `is_dicom_file` in hazenlib.__init__ please?

2. For neatness could you also move your test class `Test_is_Dicom_file` to tests/test_tools.py - you should just be able to copy and paste this onto the end of the file.

3. your tests are failing because your test function is looking for data at a path which looks like a windows filepath. Could you change the path variable so that it uses the os.join method? this way we can avoid conflicts between linux and windows paths (see below)
`E           FileNotFoundError: [Errno 2] No such file or directory: '.\\tests\\data\\tools\\dicom_yes.dcm'`

hshuaib90:
Hi @Lucrezia-Cester - the tests are still failing. Are you able to address Laurence's 3rd comment? it might solve it. thanks!

Lucrezia-Cester:
Sorry it took a while to fix the issue, changed \\ to / as Laurence suggested. @hshuaib90 

Cheers.

Issue #134, opened 2021-11-04
Improve test coverage


laurencejackson:
Our test coverage is currently at around 71% which could be better. The line numbers in the table below show which lines are currently not run during any of the tests. We should expand our existing tests or (if necessary) write new ones to cover these lines.

This is another good issue to get people familiar with the automatic testing system.
 
File | Stmts | Miss | Cover | Missing
-- | -- | -- | -- | --
hazenlib
init.py | 132 | 67 | 49% | 128, 132, 142, 147, 183–186, 191–196, 208, 211–218, 237–239, 256–258, 276–278, 286, 290–293, 297, 312–319, 324–360
exceptions.py | 21 | 4 | 81% | 17–21
ghosting.py | 156 | 50 | 68% | 24, 89–90, 94, 104–105, 132–133, 142–143, 198–226, 233–251, 255
relaxometry.py | 308 | 90 | 71% | 238–256, 630, 689–691, 745, 793–815, 833–848, 1168–1171, 1180–1183, 1195–1208, 1211–1216, 1228–1258
slice_position.py | 109 | 20 | 82% | 72–73, 98, 182, 189–202, 210, 218–219, 230
slice_width.py | 287 | 45 | 84% | 23, 26, 30, 34–39, 105, 275, 280–281, 287, 292, 355–356, 591–630, 666–669, 672–675
snr.py | 138 | 50 | 64% | 39–44, 72, 147–157, 181–191, 195–199, 231–242, 247–255, 283–294, 318, 326–328, 331
spatial_resolution.py | 290 | 99 | 66% | 84–142, 156, 238, 278–285, 313, 443–476, 497–514
tools.py | 66 | 9 | 86% | 42–49, 88, 97, 112–113
uniformity.py | 79 | 22 | 72% | 68–69, 76, 110–122, 134–136, 139, 143–146, 154
TOTAL | 1586 | 456 | 71% |

Lucrezia-Cester:
Hi @tomaroberts ! I assigned myself this issue  #90  (add tests for __init__ in hazenlib). Since you have this issue, where hazenlib __init__ is included, I was wondering if you were ok with me writing the tests for the __init__ module.

tomaroberts:
Discussed tests_snr.py with Laurence. Current tests are integration tests. Will be better to have specific unit tests for each function in snr.py.

tomaroberts:
[...]

PR #137, opened 2021-11-08
SNR centroid calculation with thresholding

### LINKED ISSUE #58: [HZN-58] SNR-get phantom shape from automatic thresholding rather than shape fitting

:::Issue recreated from Jira issue [HZN-58](https://gsttmri.atlassian.net/browse/HZN-58):::

Imported from https://bitbucket.org/gsttmri/hazen/issues/59/snr-get-phantom-shape-from-automatic  SNR routine fails if it is unable to fit a circle \(axial\) or rectangle \(sagittal or coronal\) to the image.In addition to failing if the shape fitting algorithm fails, it also assumes a cylindrical phantom and non-oblique acquisition.Using an automatic thresholding algorithm, such as skimage.filters.threshold\_li could avoid this problem and simultaneously remove the restrictions on phantom shape and acquisition plane.Preliminary investigation shows threshold\_li, threshold\_minimum and thershold\_triangle may give acceptable results, but threshold\_iosdata, threshold\_mean, threshold\_yen and threshold\_otsu are unlikely to work.Automatic thresholding should be applied to the smoothed image to reduce the effect of noise.

### LINKED ISSUE #108: SNR failing on Sola and Vida scanners
The head SNR in the sagittal plane is failing consistently across Sola and Vida.  Perhaps bright areas on the images close to the coil elements?

image

Example image data stored under `Quality/Hazen_Issues/`
Raised on behalf of David Price.

tomaroberts:
This relates to #58 and #108.

Commit implements phantom centroid calculation based on thresholding. The code tries to use the standard shape detection, then performs thresholding if this fails. Figure illustrating thresholding below:

Top left: original MR image
Top right: threshold binary mask
Bottom left: mask applied to MR image
Bottom right: different image, ie: Top Left - Bottom Left

circle_THRESHOLD

Tested in previously failing dataset (#108) which now works:

Hd_SNR_SAG_2_ORIG_24001_1

Also tested threshold method against shape detection method in a case which was previously working:

SHAPE DETECTION:
shape_SNR TRA MEAS2_10_1

THRESHOLD DETECTION:
thresh_SNR TRA MEAS2_10_1

Methods give near identical answer – centroid coordinates shifted by 1-pixel in some cases, but don't think this has significant impact on final SNR measurement.

Have also added the Vida test data (referred to in #108) to: hazen/tests/data/snr_threshold.


hazenlib/snr.py
```diff
@@ -171,32 +232,40 @@ def get_object_centre(dcm) -> (int, int):
...
+                col, row, r = shape_detector.get_shape('circle')
+            except exc.MultipleShapesError:
+                print('Warning! Found multiple circles in image, will assume largest circle is phantom.')
+                col, row, r = get_largest_circle(shape_detector.shapes['circle'])
+        else:
+            raise exc.ShapeError("Unable to identify phantom shape.")
+
+    # Threshold Detection
+    except exc.ShapeError:
+        # print('Shape Detection failed. Performing Thresholding ...')
```
laurencejackson: Can you add a logger statement that explains whats happening at the various points during this function? I think most login can be debug level but maybe add an info logging statement for when the shape detector fails and the thresholding method is used?

You can merge in (or just have a look at) origin/feature/HZN-80---adding-logging-to-screen to see how the logger will wor but basically you just put something like

`logger.info('this is an info logging statement that shows up all the time because info is the default level')`

`logging.debug('this is a debug statement that will show up when someone asked for more detailed logs')`

laurencejackson (changes requested):
Looks good, small comment below.

Can you also add test code so that the lines you've added are covered by the automated testing?

think you'll need to add a new test function - `test_threshold_image` to test/test_snr and modify another test to make sure that your exception check in line 264 is covered.

tomaroberts:
Added new testing class TestSnrThreshold – uses the VIDA data supplied by David Price which fails _hazen snr_ using conventional shape detection.

Added logging statements to get_object_centre.

As discussed, will improve the unit tests in tests_snr.py in a separate issue (#134).

laurencejackson (approved):

PR #141, opened 2021-11-09
spatial resolution vs blurring

### LINKED ISSUE #131: investigate relationship between image resolution and hazen SpatRes
Study the relationship between hazen output and images with different resolution

Lucrezia-Cester:
I applied an averaging gaussian filter to check the effect it would have on the MTF. In the commit I have inserted two images which show what happens to the MTF when the image is blurred vs when it isn't. As I was saying in the meeting, it doesn't look like the MTF changes much, although the image looks blurred.


pcw24601:
@Lucrezia-Cester, it's great that you're looking at this systematically. I'd love to get a reliable way of measuring MTF in MR.

Possibly linked issue:
[https://github.com/GSTT-CSC/hazen/issues/75](https://github.com/GSTT-CSC/hazen/issues/75)

I think there are both theoretical and practical issues with calculating MTF from MR images, and our group hasn't found an easy way to overcome them. My thoughts are below, I welcome discussion.

1. Imaging filtering. Applying filtering (some of which it can be difficult to switch off) will likely change the MTF. Two of the three test images have some sort of edge filtering applied. If you apply harsh windowing to the MTF block you can see a few pixels with zero amplitude. This clipping can also been seen in the ERF and LSF graphs.
 a. eastkent (some sort of edge enhancement filtering eliminating noise)
image
image

b. philips (looks unfiltered)
image

2. When using a magnitude image, the signal in the void is rectified. This can be seen in the ERF below, and it causes the LSF to oscillate at twice the frequency in the perspex compared to the fluid.
image

3. I'm not sure the distance from the edge is well calculated in this implementation (see linked issue above).

4. Something feels a bit funny about acquiring a Fourier transform of an image (k-space), applying the inverse FT to get the image, cropping the image then applying an FT again to get the MTF. I can't quite put my finger on it, but noise on the RF signal recorded in k-space feels like it will cause a correlation between the pixels we are looking at in a way that doesn't happen in x-ray or CT.

Lucrezia-Cester:
Hi @pcw24601, sorry for taking so long to reply. You raised very good points but I wasn't equipped with the skills to fixed this so I guess I did not work on this for a while. But I am back working on this now.

I am trying to change the current script with a simpler algorithm. 

In regard to your first point about how filtering will change the MTF, I did some research and figured out that the FT is very sensitive to deformations in the high frequency range, so filtering it will in fact change the MTF, so filtering might not be the best way to go about this problem, hence why I am trying to change the algorithm all-together. 

In regard to issue #75, you mention using a polynomial fitting curve on the ESF. I am now using a different algorithm to get the MTF and I am applying a polynomial fit as suggested in #75. I have applied the polynomial fit only to the MTF and not directly to the ESF. I think it makes some sort of sense to me as the MTF is a linear function and is "predictable" while the ESF (for our real world data) is very non-linear, which means it has a lot of singularities, which means that a polynomial fit is not the best option. But I am going to do some further research to see if it makes more sense to apply the polynomial fit directly to the ESF.

I attach a picture
Screenshot 2022-04-05 at 14 23 20
 of the smoothed MTF that I get with the new [algorithm]. Please forgive the units, I still need to figure out what they mean mathematically and then I'll properly fix those. 
This image is just to show what a smoothed
Screenshot 2022-04-05 at 14 25 34
 MTF (after I applied a polynomial fit) looks like as opposed to the non-smoothed version. 

I found a paper that deals with MTF in MRI https://pubmed.ncbi.nlm.nih.gov/8208224/. I am gonna start from here and then dig deeper into other research papers if they are available. I can't as of right now really give meaningful answers to the other points you raised but after I will do a literature review I'll try to come back to you.

Lucrezia-Cester:
Hi @laurencejackson, @pcw24601, as per the above discussion, when I used a blurring filter on the resolution images, I did not get a different value for the MTF, which I thought was very weird (this can be seen in the images that are included amongst the changed files in this PR). There were a couple of things that I thought were introducing problems with the previous script so I wrote a new script. With the new script I get different values for the MTF depending on the amount of blurring, which is what I wanted. 

without filter
Screenshot 2022-04-05 at 17 19 11





filter avg 5x5

Screenshot 2022-04-05 at 17 18 22

I need to still fix the function which gives the units in the x axis to normalise it. But I wanted to show these images because I think that they are starting to make more sense in terms of the shape of the function, which at least changes with blurring. However the oscillatory behaviour is still off, since I would expect that the response of the system is linear. Although, if I use a polynomial to fit the function so to smooth it, there is no oscillatory behaviour and it looks like a normal MTF, but I am not sure if that kind of processing is "fair" to the data.

UPDATE: to get the units of the x-axis I have used this formula freqs= np.fft.fftfreq(n, profile_length/n), which I think means n = sample size and profile_length/n is the size of the pixel. Would you agree that is how it should be done?

I am not super in tune with the physics of MRI so I welcome discussion. Cheers.

UPDATE 2: I figured out why there are oscillations. Basically it has to do with the raw data of the intensity of the image, which is not an ideal black/white edge but rather it's black then white but then the intensity drops again hence the second oscillation

Screenshot 2022-04-06 at 09 49 07

I think everything makes sense to me now. Also, you were right Paul, the polynomial fir needs to be applied to the ESF directly.

YassineAzma:
Interesting discussion. Regarding fitting functions for the edge response function, I've followed the results acquired in this 
[paper](https://github.com/GSTT-CSC/hazen/files/8424983/ComparisonofdifferentanalyticaledgespreadfunctionmodelsforMTFcalculationusingcurve-fitting.1.pdf) where they compare different analytical functions for fitting the ERF. They recommend a logistic function. I've generalised this to a sigmoid function of the following form:

ERF = a + b/(1+exp(c*(x-d)))^e

Regarding the spatial frequencies on the x-axis, this is what I've done:

1) Create an integer array from -N/2:N/2 (for even lengths), -(N-1)/2:(N-1)/2 (for odd lengths) where N is the length of your LSF
2) Define your spatial frequency sampling rate as (1/bin size). This would be 1/pixel spacing in a particular direction if you have not supersampled. If you have, it would be (1/(pixelspacing*(1/samplefactor)).
3) Multiple Step 1 by Step 2 and divide by N to get the spatial frequency per bin (i.e. your x-axis)

I've started working on this again for the ACR phantom using this [method](https://opg.optica.org/oe/fulltext.cfm?uri=oe-22-5-6040&id=281325). It's very similar to the current hazen implementation, though the edge is acquired in super-resolution to bypass Nyquist limits, and the edge slope itself is determined using 2D surface fitting to avoid the noise sensitivity that arises when using line profiles. I'm quite curious as to whether I can replicate your result. Since we have the "newer" model of the ACR phantom it comes with a distortion board, which is in place of the grid on slice 5. With this, I can do an MTF in both directions. NB: You'll need to click on the images in order to see them clearly.

**Without filter**
no blur ap
no blur lr

**2D Gaussian filter with sigma_x = 0.5 and sigma_y = 0.2 and a kernel size of [3,3]**
blur ap
blur lr

**2D Gaussian filter with sigma = 0.5 and a kernel size of [5,5]**
square blur ap
square blur lr

Whilst my MTF50 values are changing as expected, I don't get that oscillatory behaviour you're seeing in your MTF implementation. This does come with the caveat that I apply a (peak-centred in the future) Hamming window to the LSF to suppress Gibbs ripple.

pcw24601:
@Lucrezia-Cester @YassineRMH This is really interesting work. I only discovered MTF about 3 years ago (I'm fairly new to medical physics) and thought 'I can apply this to our MR images'. My colleagues more or less told me 'We've tried, it doesn't really work for MR, but knock yourself out'. I couldn't get it to produce answers I was happy with and have now passed through that rite of passage, am a bit greyer and (maybe) less arrogant! I'd still love to see this work.

### 1. Fitting ERF / MTF
@Lucrezia-Cester -- the reference for fitting the ERF instead of the MTF is from IPEM report 112 [Quality Control and Artefacts in Magnetic Resonance Imaging](https://www.ipem.ac.uk/resources/books/report-112-quality-control-and-artefacts-in-magnetic-resonance-imaging/):

image

I haven't reviewed your updated code, but when I first looked at the Hazen code, the x-position was not calculated properly. If you use a slanted edge the x coordinate should refer to the distance from the edge. Since the edge is slanted, this will be different for each image pixel and in general the x-coordinates will not be evenly spaced. Conventional FFT will not work (although I perhaps you could FFT each profile line then combine at the MTF level--maybe that's what you did?).

@YassineRMH's sigmoid ERF fits look really close to the edge response. There is a slight downward trend in raw data on the 'hot' side of the edge; from the images, this may be caused by the coil sensitivity variation. Using prescan normalisation may reduce this effect (or maybe model it in your fit).

### 2. Smoothing filters
The smoothing I was concerned about was happening during image reconstruction on the scanner. It is often difficult, and sometimes (almost) impossible to disable this during acquisition, and the scanner-filtered images are all we have to run through our analysis. A side question is: What are we trying to measure? Something as close as possible to what the scanner acquires, or something that reflects the final images produced in a clinical environment (with all the filters on)? My feeling is we need both, as a problem with the former indicates probable hardware degradation, even if clever post processing on the scanner can mask these effects; whilst the latter could also impact patient pathway due to missed diagnoses.

The default magnitude images are also a problem and signed/real valued/complex images may provide a better solution to the rectified signal in the void, but this would require changes to acquisition, and some of the signed valued images we've acquired have had artefacts not visible in the magnitude images. (The need for complex images is mentioned in the abstract to the paper @Lucrezia-Cester referenced.) Do you think is it possible just to analyse the 'hot' side of the edge (maybe rotate the image to simulate a signed version of the cold side)?

### 3. Gibbs ringing
How should we deal with Gibbs ringing? It is a deterministic part of the image (rather than random noise) and it feels that any filtering to remove it will likely reduce the MTF we're trying to measure. @YassineRMH--did you produce images without the Hamming window filter?

YassineAzma:
@pcw24601 I must admit that I'm still very leery of MTF in MRI. Much of what I'm doing in my code feels more akin to data sanitisation than modelling any real physical processes. When I was a trainee, my supervisor, Maria Schmidt, was very down on MTF in MRI, having done some work with Sian Curtis from Bristol looking into it. I don't quite recall her misgivings, so I'll get in touch and see what she thinks. My current version works in that the values make sense comparatively (sensitive to differences in base resolution, phase/base resolution asymmetry, addition of filters), but without a gold standard I can only say it's semi-quantitative at best. I'll have to consider how I would model the intensity variation in the image seen without pre-scan normalization. 

The motivation for me working on this again is that we have one scanner (Siemens Sola 1.5T) that still has its inline filter (roll-off filter applied to k-space) enabled, and we were interested in quantifying the degree of blurring it caused. One of my colleagues used my script and put together a short report. 

[Inline_filter_quick_summary.pdf](https://github.com/GSTT-CSC/hazen/files/8427008/Inline_filter_quick_summary.pdf)

It's since been turned off, but the effect seems to be subtle to the point where my script was unable to detect any differences. It has to be mentioned that the phantom/ROI location is not exactly the same, as it was necessary to return the couch to home position between sequences. The Sola only allows for inline filters to be changed via restart of the scanner.

I agree that data should be acquired both with filters that would typically be on during clinical scanning and without any. This way intrinsic system performance can be assessed (allowing for comparison between scanners) and we get a result that is reflective of what would be output in a clinical scan. I have little to add to offer about the potential of working with signed data. I've only worked with the phase and magnitude data, not the individual real/imaginary channels. I don't quite understand the purpose of rotating the complex image. Could you elaborate further?

The Hamming window is applied directly to the LSF, not to the images. I've changed the output ROI location from my edge detection script to be away from those hyperintense targets you can see in the hypo-intense central square. They were causing that upward drift in the "cold" part of the ERF.

**With the Hamming window**
with hamming lr
with hamming ap

**With the Hamming window cubed - this makes the window steeper (and more like a Lorentzian)**
with hamming cubed lr
with hamming cubed ap

**Without the Hamming window**
without hamming lr
without hamming ap

You can see that the effect is quite subtle (but still undesirable), though I think it would preserve more of the peak if centred there, rather than MATLAB's default behaviour of the centre of the array.

Lucrezia-Cester:
Hi @laurencejackson, the updated SR script is now ready for review.

pcw24601:
Sorry for the delay in replying.

# Sola inline filter?
@YassineRMH--where is the 'inline filter' option that needs a reboot on the Sola? Is this different from the 'raw' and 'elliptical' filter options in the Resolution -> Filter card?

# Rotating the signed / complex image
My understanding (for Siemens) is that their 'real' data isn't the real part of the complex signal in mathematical terms, but rather the magnitude multiplied by (-1) if the phase is more than 180 degrees. I referred to this as 'signed', sorry for the confusion (which may well be in my understanding).

The 'rotating' I was picturing was a thought to address the rectification in the magnitude image. The problem is the region in the solid red circle is rectified--alternate oscillations should be negative:
image

If all we have is the magnitude image, my thought is to discard the cold side and replace it with a rotated version of the hot side (green dashed rectangle) which does not suffer from the rectification issue:
image

# Analytical differentiation
@Lucrezia-Cester, I had a quick look at your code and it looks like a bit improvement! Can I check--is it numerically differentiating the ERF to get the LSF (line 68)? Did you consider doing this analytically from the polynomial coefficients ( if f(x) = a_(n) * x^n + a_(n-1) * x^(n-1) + ... then f'(x) = n*a_(n) * x^(n-1) + (n-1) * a_(n-1) * x^(n-2) + ...; the a_0 will disappear). Also, did you need to go as high as 20 coefficients?

YassineAzma:
@pcw24601 

I'm just as guilty of being late to reply!

Those filters you're referring to are sequence-specific. There is an additional filter parameter that is hidden in the global settings. Siemens refers to it in their documentation as both an implicit image filter and an inline image filter. You can access it through the Administration Portal (with service key required) in Technical Configuration -> MRI Configuration -> Measurement Configuration.

Thanks for the clarification. Given Gibbs is, as you mentioned previously, deterministic, wouldn't it be more mathematically "correct" to instead consider we have an absolute of a sinusoid there and just reconstruct the actual ripples. We would expect, in the absence of sensitivity effects, that the odd oscillations would be of negative polarity. Happy to explore this at some point!

pcw24601:
@YassineRMH 
> Thanks for the clarification. Given Gibbs is, as you mentioned previously, deterministic, wouldn't it be more mathematically "correct" to instead consider we have an absolute of a sinusoid there and just reconstruct the actual ripples. We would expect, in the absence of sensitivity effects, that the odd oscillations would be of negative polarity. Happy to explore this at some point!

Great idea, certainly interesting to investigate. Things might become tricky once the noise becomes comparable to the signal and we start getting Rician noise problems. In the example we were looking at, once you get above ~85 in the signal looks like its in the noise (edge is about 60, about 25 pixels from the edge). On the hot side, the oscillations are still visible when you hit the origin, so a range of 60.

I simulated some noisy data in Excel. `signed_noisy = exp(-t/t0) * sin(2pi*t) + GaussianNoise` and `mag_noisy = abs(signed_noisy)`. To my eye, the oscillations vanish into the noise in `mag_noisy` long before they do in `signed_noisy `.

image

It also might be more challenging when the oscillations aren't as well defined as in this example (cf the other one I pasted on 10 Nov).

What do you think?

pcw24601:
> There is an additional filter parameter that is hidden in the global settings. Siemens refers to it in their documentation as both an implicit image filter and an inline image filter

I've been chatting with my colleague about this some more. If you select 'Pre-scan normalisation' then 'save unfiltered' (or apply 2DDC + save unfiltered), does the unfiltered set of images bypass the in-line filter?

YassineAzma:
> > There is an additional filter parameter that is hidden in the global settings. Siemens refers to it in their documentation as both an implicit image filter and an inline image filter
> 
> I've been chatting with my colleague about this some more. If you select 'Pre-scan normalisation' then 'save unfiltered' (or apply 2DDC + save unfiltered), does the unfiltered set of images bypass the in-line filter?

@pcw24601 I don't think so. My understanding is that the Unfiltered checkbox provides you with a dataset both with and without the associated filter the checkbox is linked to. These are referred to by Siemens as explicit image filters. There's also similar setups for post-processing like Dynamic Field Correction in DWI.

image

image

The implicit image filter is always there, though not on every sequence as mentioned above.

tomaroberts:
Closing this PR as it is superseded by #286

PR #143, opened 2021-11-15
Porting hazen to python 3.9

### LINKED ISSUE #142: Update hazen to work with Python 3.9
hazen is currently compiled with Python 3.6. This is end-of-life at the end of 2021. We should keep hazen up-to-date with the latest Python versions.

Further still, installing Python 3.6 is particularly troublesome for M1 Apple Silicon users. There are ways of running Python 3.9 on M1 chips via Rosetta (https://github.com/pyenv/pyenv/issues/1768#issuecomment-843361418), but this is cumbersome.

### LINKED ISSUE #158: Use Ibn Al-Haytham's image instead of CSC logo
**Is your feature request related to a problem? Please describe.**
I think it would be more appropriate to have the image of Ibn Al-Haytham (who inspired the name Hazen) instead of the CSC logo. I think it would be better for community to development to not include CSC branding.

**Describe the solution you'd like**
Replace CSC logo with Ibn Al-Haytham's image in README.

**Describe alternatives you've considered**
NA

**Additional context**
NA

tomaroberts:
See #142.

Currently, for python3.9, hazen requires conda package management as not all wheels working in pip. hazen can be run in 3.9, but some tools failing because of changes to old libraries, and possibly other issues.

TODO:
- Fix cv2 related error. OpenCV has changed syntax. Now recommends working on binary image prior to findContours. Might affect our results? See here: https://docs.opencv.org/4.x/d4/d73/tutorial_py_contours_begin.html
- others...


tomaroberts:
Reminder: OpenCV findContours() changed the number of output arguments between versions... Unhelpful -_-

https://stackoverflow.com/a/58244568

tomaroberts:
tests/resolution/RESOLUTION passes
tests/resolution/eastkent fails
tests/resolution/philips fails

eastkent and philips failing here:

image

These 5 all have rescale_to_byte(), thresh() and find_square() in common. I'm pretty sure find_square is working, otherwise the RESOLUTION test set wouldn't pass.

In the eastkent/philips test_data, I think something might be happening with rescale_to_byte() or thresh() which is causing find_square() to output coordinates which don't match the test values. e.g.:

image

I've also noticed that there is a subtle different in the images output between hazen3.6 and hazen3.9:

Left = old hazen (3.6)
Right = new hazen (3.9)

image

The 3.9 images are smoother when loaded in – are they supposed to look like left or right?! Is the 'shimmer' in the 3.6 images an MR artefact or python artefact?

tomaroberts:
Should add, current spatial_resolution results:

RESOLUTION
perfect match between py3.6/py3.9

eastkent
py3.6: FE = 0.98 / PE = 0.99
py3.9: FE = 0.99 / PE = 0.99

philips:
py3.6: FE = 0.49 / PE = 0.54
py3.9: FE = 0.49 / PE = 0.49

So pretty similar, seemingly indicating numerical/subtle threshold or contrast issue.

environment.yml
```diff
@@ -0,0 +1,47 @@
+name: hazen-3.9-miniforge
```
hshuaib90, line 1: What does this file do?

hazenlib/spatial_resolution.py
```diff
@@ -174,9 +174,9 @@ def thresh_image(img, bound=150):
 
 
 def find_square(img):
-    cnts = cv.findContours(img.copy(), cv.RETR_LIST, cv.CHAIN_APPROX_SIMPLE)
+    cnts = cv.findContours(img.copy(), cv.RETR_LIST, cv.CHAIN_APPROX_SIMPLE)[0]
```
hshuaib90, line 177: Presumably this is to take into account the different OpenCV version? But you haven't updated the pip requirements version here (unless that's what environment.yml is doing??)

hshuaib90:
> tests/resolution/RESOLUTION passes tests/resolution/eastkent fails tests/resolution/philips fails
> 
> eastkent and philips failing here:
> 
> image
> 
> These 5 all have rescale_to_byte(), thresh() and find_square() in common. I'm pretty sure find_square is working, otherwise the RESOLUTION test set wouldn't pass.

I just checked and all the resolution tests are failing with your recent commit #866dc22 . `find_square` is the only code you've changed from develop branch according to the GitHub PR.

> 
> The 3.9 images are smoother when loaded in – are they supposed to look like left or right?! Is the 'shimmer' in the 3.6 images an MR artefact or python artefact?


I'm not sure why the appearance of the images changes between Python versions, but it is worrying. Need to see if anyone else can reproduce locally.

environment.yml
tomaroberts: Sorry, shouldn't have spread between two places. Bit more context in #142.

Basically, installing some of our libraries (mainly scikit-image) doesnt work on M1 silicon via pip. So instead I made the venv in miniforge (just a smaller version of conda). The environment.yml is the equivalent to our normal requirements.txt.

I specifically made most of it install via pip anyway, so it was easy to see which ones work via pip and which only work via conda.

hazenlib/spatial_resolution.py
tomaroberts: Correct – see comment about on .yml file.

tomaroberts:
> I just checked and all the resolution tests are failing with your recent commit #866dc22 . `find_square` is the only code you've changed from develop branch according to the GitHub PR.

image

These are the latest results I'm getting from running `pytest tests/`

environment.yml
hshuaib90: Sorry for being a bit slow here. But doesn't the dockerisation of Hazen make this issue a moot point?

hshuaib90: Sorry I should have made this comment on the issue when you first raised it.

tomaroberts:
I've just noticed the Git Actions pytest section where it's failing all the spatial_resolution tests, as you say.

I think the output is different to the tests I'm running locally on my laptop because the Git Actions test is building the code from requirements.txt with py3.6 and using the old OpenCV version, meaning that my alteration to findContours() is causing it to fail.

tomaroberts:
@hshuaib90 
> I'm not sure why the appearance of the images changes between Python versions, but it is worrying. Need to see if anyone else can reproduce locally.

Have discovered that the matplotlib version makes a significant different on how the images are displayed...!

See images below, generated using identical python3.9 venvs, only difference is matplotlib versions:

**Top:** matplotlib==3.1.3 (which is used in main hazen version)
**Bottom:** matplotlib==3.4.3

Screenshot 2021-11-18 at 14 12 49

Matplotlib shouldn't be affecting the values in the underlying dicom pixel_arrays however as it's just for plotting, so the discrepancies in hazen output values is something else.

At the moment, I'm pretty sure that the hazen code is behaving consistently between the 3.6 build and my 3.9-miniforge build, but a subtle change in one of the libraries is giving us slightly different results, hence the tests are failing.

My feeling is that we just need to update the tests to reflect these slight changes in values.

[environment-mplib3.1.3.yml.txt](https://github.com/GSTT-CSC/hazen/files/7563541/environment-mplib3.1.3.yml.txt)

tomaroberts:
This is now ready for review.

Below is a summary of all the changes to this point in getting hazen into python3.9, given that this PR is a bit long already:

### Why python3.9?

- Important to keep software up-to-date (eliminate that technical debt...)
- At the time of writing, M1 silicon Macbooks (which we use in CSC) have numerous incompatibilities with older python packages, including python itself, scikit-image, numpy, etc. Many of these were used in the previous hazen build.

### Changelog

- **Installation requires conda _if using M1 Mac_**
	- NB: this is a temporary measure until various packages are updated to install via pip on M1 chips.
	- I will update installation documentation soon with conda installation guide
	- Could make a requirements.txt and get someone to test pip installation on a non-M1 computer.
-  **Updated dependencies**
	- these are now in `environment.yml` which conda uses, rather than `requirements.txt`
	- Installed as many as possible through pip so easier to transition away from conda in the future
		- Main changes:
			- python=3.9.7
			- opencv-python==4.5.4.58
			- scikit-image=0.18.3
			- matplotlib==3.4.3
			- scipy==1.7.2
			- psycopg2-binary==2.8.4
			- numpy==1.2.1
			- +some others
- **Changed .github/workflows/tests_development.yml**
	- Replaces older python3.6 version
	- Uses conda
- **OpenCV version change**
	- Caused a lot of issues
	- `findContours()` -- number of arguments output changes between versions
	- Main problem identified with `cv.minAreaRect()` function
		- In v4.5 the output arguments have changed
			- Seems the rectangle orientation is defined differently
			- `size` output has swapped around, so x=y and y=x
			- `angle` 90 degrees different than in OpenCV v3.4
			- See comparison below. Both cases running: `hazen spatial_resolution tests/data/resolution/philips`

              **hazen3.6**
              minRectArea3 6
              rect[1]=(79,80)
              rect[2]=-10.9
              
              **hazen3.9**
              minRectArea3 9
              rect[1]=(80,79)
              rect[2]=79.1 

- **matplotlib version change**
	- Updating version changes the appearance of displayed images! See [here](https://github.com/GSTT-CSC/hazen/pull/143#issuecomment-972910498).
	- Doesn't affect underlying values, but explains why `hazen --report` output images look slightly different

### Testing

Implementing these changes caused various tests to fail. These boiled down to 3 sets of related failures:

 **1. test_tools**
 
tools_test_fails

The coordinates of the generated rectangles were failing assertion tests because `minAreaRect()` was effectively rotating shapes 90 degrees compared to old hazen. Changing the size/angle outputs fixed these (along with rounding to 2 decimal places).
 
 2. **test_spatial_resolution**
 
spatial_res_test_fails

Updating the code to reflect the changes due to `minAreaRect()` helped with this one, but didn't fully fix the tests. The square/rectangle shape tests were still failing because the coordinates of the hazen3.9 generated shapes were one or two pixels different to the shapes in the test code.

I checked the various OpenCV functions involved with detecting and generating the rectangle shapes, such as cv.findContours(), cv.arcLength(), cv.approxPolyDP() and the outputs of these are the same between 3.6 and 3.9 builds. Only minAreaRect() had an obvious difference.

I think there is some small numerical or algorithmic difference in the updated version of OpenCV which has ever so slightly altered the rectangular shape generation. The images output from `hazen spatial_resolution --report` all look sensible in hazen3.9, and basically the same as hazen3.6. Plus, the final spatial_resolution output values are only marginally different, which gives confidence the underlying functionality is the same.

I've updated the test_spatial_resolution.py values to reflect the numbers generated by hazen3.9. Again, shifts in coordinate positions of a few pixels at most.
 
 3. **test_models**
 
models_test_fails

Simple fix, just required updating werkzeug from 0.15.4 to 0.15.6.

README.md
```diff
@@ -25,12 +25,81 @@ Quality assurance framework for Magnetic Resonance Imaging
 ## Overview
 
 ---
+hazen is a software framework for performing automated analysis of MRI Quality Assurance data.
 
+Some examples of what hazen can perform:
 
+| SNR measurements   | Spatial Resolution measurements |
+| ------------------ | ------------------------------- |
+| _insert snr image_ | _insert spatial res image_      |
```
hshuaib90: @tomaroberts can you add something here to just remove the examples section?

README.md
```diff
@@ -25,12 +25,81 @@ Quality assurance framework for Magnetic Resonance Imaging
...
+First, clone this repo, then follow the instructions for your operating system. To clone:
+```bash
+git clone git@github.com:GSTT-CSC/hazen.git
+```
+
+#### MacOS
+
+```bash
+# Download and install conda (we have tested with both Miniforge and Miniconda):
+brew install miniforge
```
hshuaib90: Can you clarify what non-m1 macs need to please.

tests/test_tools.py
```diff
@@ -63,18 +63,36 @@ def test_sag_rectangle(self):
         arr = pydicom.read_file(self.SAG_RECTANGLE_PHANTOM_FILE).pixel_array
         shape_detector = hazen_tools.ShapeDetector(arr=arr)
         centre, size, angle = shape_detector.get_shape('rectangle')
+        centre = hazen_tools.round_tuple(centre)
```
hshuaib90: Why is this round_tuple necessary? I think it introduces more risk than it's worth. If the "true" answer has too much precision, then we can do an assert almost_equal instead.

hazenlib/slice_width.py
```diff
@@ -362,7 +362,7 @@ def get_initial_trapezoid_fit_and_coefficients(profile, slice_thickness):
         n_ramp = 47
         n_plateau = 55
 
-    trapezoid_centre = round(np.median(np.argwhere(profile < np.mean(profile)))).astype(int)
+    trapezoid_centre = int(round(np.median(np.argwhere(profile < np.mean(profile)))))
```
hshuaib90, line 365: Why is this change necessary?

hshuaib90 (changes requested):

README.md
tomaroberts: These were placeholders to remind me for when I overhaul the README.md

README.md
tomaroberts: Yeah.

@laurencejackson – can you do a test install on your Intel Macbook? Using the requirements.txt.

tests/test_tools.py
tomaroberts: Ah, wasn't familiar with almost_equal. Will work that in instead. Thanks.

hazenlib/slice_width.py
tomaroberts: This can be removed actually. Was an early change when I thought a numpy version change was causing problems.

tomaroberts: Resolved by 2ca01a3

README.md
laurencejackson: What do people think of renaming this section to MacOS (M1) and then renaming "Linux" to "Linux and Mac OS (Intel)" since the instruction for intel macs and Linux will be identical. 

This might not be ideal since it gives users two separate install instructions for macs, but I'd rather users and developers didn't install conda unless they have to, and us using it here is intended to be temporary anyway. 

Since most users will be using the dockerised version anyway I think we should have a  section under install called "docker", which should be first in the list. Otherwise people won't see the docker instructions until they get down to 'usage". Could we add a paragraph saying something along the lines of "the recommended way to use hazen is through docker" too?

README.md
tomaroberts: @laurencejackson @hshuaib90 

I've temporarily made a separate branch for the README edits so as not to interfere with this PR. You can view latest version here: https://github.com/GSTT-CSC/hazen/tree/readme-dev

Also includes re-jigging of usage instructions, i.e.: Docker first, then Mac/Linux, then M1 users.

PS: I've written it assuming pip install works on Linux/Intel Macs (just awaiting Laurence's confirmation)

tomaroberts:
Updated requirements.txt to specify mistune version as m2r install was causing error. See: https://github.com/miyakogi/m2r/issues/66

tests/test_tools.py
tomaroberts: Fixed with commit #fc2f215

README.md
laurencejackson: Tested on python 3.9.7 on intel mac and everything works as expected following the readme you linked

.github/workflows/cli-test.yml
```diff
@@ -15,10 +15,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Set up Python 3.6
+    - name: Set up Python 3.9
       uses: actions/setup-python@v2
       with:
-        python-version: 3.6
+        python-version: 3.9.7
```
laurencejackson: can we remove the patch version so it's just 3.7- so we always test on the latest version.

.github/workflows/tests_development.yml
```diff
@@ -11,6 +11,9 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    defaults:
```
laurencejackson, line 18: Could we also try adding a basic platform strategy as described [here](https://github.com/actions/setup-python)? I think for now something like the below will be fine, so we test the two main OSs which we know are used and the current and future major python versions.

```yaml
runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest]
        python-version: ['3.9', '3.10']
```

.github/workflows/tests_development.yml
```diff
@@ -30,20 +33,27 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
-
     steps:
     - uses: actions/checkout@v2
 
-    - name: Set up Python 3.6
-      uses: actions/setup-python@v2
+    - name: Set up Conda
```
laurencejackson: Since the conda install method is a temporary workaround I would prefer us to leave the test_development.yml workflow as it is (but using python 3.9 or the matrix above) and install and test using pip. 

But we definitely need to have tests covering the conda install method while it is in use so could you create a new job in this workflow `test_conda` that installs with conda. This way once the pip wheels for the necessary packages are updated to work with arm64 we can easily disable that workflow without needing to modify this one.

.github/workflows/tests_release.yml
```diff
@@ -35,10 +35,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Set up Python 3.6
+    - name: Set up Python 3.9
```
laurencejackson: we will need a new job `test_conda` in here too

laurencejackson (changes requested):

tomaroberts:
December release of [scikit-image added arm64 support](https://github.com/scikit-image/scikit-image/blob/main/doc/release/release_0.19.rst) so hazen install via pip now works on M1 Macs \o/

Thanks to @laurencejackson for additional assistance in OpenSSL installation, required for psycopg2-binary package and killing conda.

PR #153, opened 2021-11-24
wrote new function to find LSF

### LINKED ISSUE #71: [HZN-71] SpRes-mavis_deriv function problems

:::Issue recreated from Jira issue [HZN-71](https://gsttmri.atlassian.net/browse/HZN-71):::

Imported from https://bitbucket.org/gsttmri/hazen/issues/72/spres-mavis_deriv-function-problems  1. The calculated gradient \(L22\) is double the actual gradient. This can be shown with `mavis_deriv(_,range(10)) `which returns an array of 2.0s, rather than 1.0s. As the MTF is normalised this bug does not effect the final return value.2. The docstring does not accurately reflect the function’s behaviour.3. The function takes five input arguments but only uses one of them \(`a`\). Consider removing unused variables.4. Consider adding a reference to any published literature etc. which recommends this method of differentiation over e.g. standard Newton difference quotient [https://en.wikipedia.org/wiki/Numerical\_differentiation](https://en.wikipedia.org/wiki/Numerical_differentiation)‌

Lucrezia-Cester:
As mentioned, I could not find the literature to support the Maivis equation which finds the LSF. However it seems the LSF can be simply calculated by taking the derivative of the ESF http://sciencewise.info/resource/line_spread_function/Line_spread_function_by_Wikipedia. So I changed the function directly. Let me know if this fix is ok. 
The calculate_mtf_test is not failing but I think it's because the output will be different with the new function I added.

I attach a picture to show the difference in the appearance of the MTF 
maivis_vs_deriv

In this paper https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3643984/ they also claim they find LSF from differentiation of ESF, which supports this new change of function.

### DIFF
--- a/hazenlib/spatial_resolution.py
+++ b/hazenlib/spatial_resolution.py
@@ -71,6 +71,7 @@ def maivis_deriv(x, a, h=1, n=1, axis=-1):
     return b
 
 def deri(a):
+    # This function calculated the LSF by taking the derivative of the ESF. Reference: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3643984/
     b = np.gradient(a)
     return b
 
--- a/hazenlib/spatial_resolution.py
+++ b/hazenlib/spatial_resolution.py
@@ -70,6 +70,9 @@ def maivis_deriv(x, a, h=1, n=1, axis=-1):
     b.extend([b[-1]] * 3)
     return b
 
+def deri(a):
+    b = np.gradient(a)
+    return b
 
 def create_line_iterator(P1, P2, img):
     """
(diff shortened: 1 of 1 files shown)


laurencejackson:
Hi Lu, this looks good. Thanks for doing the research to ensure this change is correct. Looking at the screenshots your posted in #71 it looks like the results are good. 

Could you have a look at why the tests are failing now? I imagine it will be due to the result being different with this new method. 

Could you also address the numbered points made in the original issue? copied below:

1.  The calculated gradient (L22) is double the actual gradient. This can be shown with mavis_deriv(_,range(10))which returns an array of 2.0s, rather than 1.0s. As the MTF is normalised this bug does not effect the final return value.
2.    The docstring does not accurately reflect the function’s behaviour.
3.    The function takes five input arguments but only uses one of them (a). Consider removing unused variables.
4.    Consider adding a reference to any published literature etc. which recommends this method of differentiation over e.g. standard Newton difference quotient https://en.wikipedia.org/wiki/Numerical_differentiation

I think since we're looking at removing the old `maivis_deriv` function entirely you can just delete it, which would address 
points 2 and 3.

Lucrezia-Cester:
Hi @laurencejackson, in response to the previous comment:
 
1) I tried and the gradient is correctly calculated as 1 with the new function
2)Not relevant anymore (although I added a short docstring)
3)Not relevant anymore
4) I added the reference.

tomaroberts:
Closing PR as duplicate of #191 which supersedes this.

PR #178, opened 2021-12-14
Fix docker cli args - requires new hazen-app in path!

laurencejackson:
should now be able to use --version and -h options in CLI as well as -b build tag


tomaroberts (approved):
Tested locally on my laptop – working as intended.

PR #185, opened 2021-12-15
Updated ghosting function – fixes sampling point issue

### LINKED ISSUE #51: [HZN-51] Prevent ghosting from returning negative values

:::Issue recreated from Jira issue [HZN-51](https://gsttmri.atlassian.net/browse/HZN-51):::

*strong text*Imported from https://bitbucket.org/gsttmri/hazen/issues/52/prevent-ghosting-from-returning-negative

### LINKED ISSUE #99: [HZN-99] Hazen produced ghosting values do not match ImageJ values
Transferred by @LaurenceJackson on Feb 24, 2021, 16:31

:::Issue recreated from Jira issue [HZN-99](https://gsttmri.atlassian.net/browse/HZN-99):::

Hazen ghosting was used to analyse images from ECH Vida acceptance testing.  

The values produced by hazen did not match those produced with ImageJ, and did not match the visible ghosting seen on the images. Values produced either underestimated ghosting or over estimated (for TE=120ms, NSA=1). 

Hazen produced values and ghosting attached below.

### LINKED ISSUE #109: Ghosting measurements on Philips Ingenia 1.5T
Ghosting not working- according to position of max ghosting not in rectangular box as desired.  Perhaps something in Philips header?

image

Example image data stored under `quality/Hazen_Issues/`
Raised on behalf of David Price.

### LINKED ISSUE #160: Ghosting script does not work for physicists
ghosting_issue

Lucrezia-Cester:
Hello @laurencejackson, @tomaroberts  , I have fixed the ghosting script. The biggest problem was that the ghosting slice was always placed at random points in the image, not where it should have been, namely int the phase encoding direction. There is now a new function called get_ghosting_slice2 which selects the correct ghosting slice. I removed the old get_ghosting_slice function. 
I have had to change the code slightly in the main function and in the report_path if statement because it didn't plot correctly. 
Finally I changed the rescale_to_byte function because the original one changes the values and spreads them into a more widened histogram. I can put this back how it was but I think ideally that function will need to be changed at some point for all the hazenlib scripts.

### DIFF
--- a/hazenlib/ghosting.py
+++ b/hazenlib/ghosting.py
@@ -211,7 +211,7 @@ def get_ghosting(dcm, report_path=False) -> dict:
 
     ghosting = calculate_ghost_intensity(ghost, phantom, noise)
 
-    if report_path:
+    if report_path == True:
         import matplotlib.pyplot as plt
         fig, ax = plt.subplots()
         x1, x2, y1, y2 = bbox
--- a/hazenlib/ghosting.py
+++ b/hazenlib/ghosting.py
@@ -169,7 +169,7 @@ def get_eligible_area(signal_bounding_box, dcm, slice_radius=5):
 
 
 
-def get_ghost_slice2(signal_bounding_box, dcm, slice_radius=5):
+def get_ghost_slice(signal_bounding_box, dcm, slice_radius=5):
     eligible_area = get_eligible_area(signal_bounding_box, dcm, slice_radius=slice_radius)
     ghost_slice = np.array(
         range(min(eligible_area[1]),max(eligible_area[1])), dtype=np.intp)[:, np.newaxis], np.array(
@@ -195,7 +195,7 @@ def get_ghosting(dcm, report_path=False) -> dict:
 
     signal_centre = [(bbox[0]+bbox[1])//2, (bbox[2]+bbox[3])//2]
     background_rois = get_background_rois(dcm, signal_centre)
-    ghost_col, ghost_row = get_ghost_slice2(bbox, dcm, slice_radius=slice_radius)
+    ghost_col, ghost_row = get_ghost_slice(bbox, dcm, slice_radius=slice_radius)
     ghost = dcm.pixel_array[(ghost_col, ghost_row)]
     signal_col, signal_row = get_signal_slice(bbox, slice_radius=slice_radius)
     phantom = dcm.pixel_array[(signal_row, signal_col)]
--- a/hazenlib/ghosting.py
+++ b/hazenlib/ghosting.py
@@ -45,6 +45,7 @@ def get_signal_bounding_box(array: np.ndarray):
         if voxel > signal_limit:
             signal.append(idx)
 
+
     signal_column = sorted([voxel[1] for voxel in signal])
     signal_row = sorted([voxel[0] for voxel in signal])
 
@@ -52,7 +53,6 @@ def get_signal_bounding_box(array: np.ndarray):
     lower_row = max(signal_row)
     left_column = min(signal_column)
     right_column = max(signal_column)
-
     return left_column, right_column, upper_row, lower_row,
 
 
@@ -111,6 +111,7 @@ def get_background_slices(background_rois, slice_radius=5):
     slices = [(np.array(range(roi[0]-slice_radius, roi[0]+slice_radius), dtype=np.intp)[:, np.newaxis], np.array(
         range(roi[1]-slice_radius, roi[1]+slice_radius), dtype=np.intp))for roi in background_rois]
 
+
     return slices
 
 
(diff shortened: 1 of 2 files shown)


Lucrezia-Cester:
Hi @laurencejackson , @tomaroberts I fixed the ghosting script. I also fixed the tests.

Lucrezia-Cester:
I have also fixed the issue which said: prevent ghosting from outputting negative values. That issue is linked. For that I only had to add an absolute value in the ghosting formula. @tomaroberts @laurencejackson

hazenlib/ghosting.py
```diff
@@ -223,25 +244,28 @@ def get_ghosting(dcm, plotting=False) -> dict:
         img = cv.rectangle(img.copy(), (x1, y1), (x2, y2), (255, 0, 0), 1)
 
         ax.imshow(img)
+        fig.savefig(f'{report_path}.png')
```
tomaroberts, line 247: When I run `hazen ghosting ... --report` it outputs a file called `True.png`. I think it's because you have used `report_path` here, which = True or False.

tomaroberts, line 247: When I run `hazen ghosting ... --report` it outputs a file called `True.png`. I think it's because you have used `report_path` here, which = True or False.

On a related note, the function always outputs `True.png`, even without the `--report` option. Is that supposed to happen?

hazenlib/ghosting.py
```diff
@@ -131,75 +132,95 @@ def get_eligible_area(signal_bounding_box, dcm, slice_radius=5):
...
     x, y = hazenlib.get_pixel_size(dcm)  # assume square pixels i.e. x=y
     # ROIs need to be 10mmx10mm
     slice_radius = int(10//(2*x))
 
+
     signal_centre = [(bbox[0]+bbox[1])//2, (bbox[2]+bbox[3])//2]
     background_rois = get_background_rois(dcm, signal_centre)
     ghost_col, ghost_row = get_ghost_slice(bbox, dcm, slice_radius=slice_radius)
-    ghost = dcm.pixel_array[(ghost_row, ghost_col)]
+    ghost = dcm.pixel_array[(ghost_col, ghost_row)]
```
tomaroberts, line 199: Is this an intentional change, swapping around ghost_col / ghost_row? Just checking in case you left if that way round during testing.

hazenlib/ghosting.py
```diff
@@ -131,75 +132,95 @@ def get_eligible_area(signal_bounding_box, dcm, slice_radius=5):
...
         import matplotlib.pyplot as plt
         fig, ax = plt.subplots()
         x1, x2, y1, y2 = bbox
 
-        img = hazenlib.rescale_to_byte(dcm.pixel_array)
+        img = dcm.pixel_array
+        img = img.astype('float64')
+        #print('this is img',img)
+        img *= 255.0 / img.max()
+        #img = hazenlib.rescale_to_byte(dcm.pixel_array)
```
tomaroberts, line 223: The `rescale_to_byte` function is also used in `spatial_resolution.py` and possibly others. I'm a little concerned by changing it here and not doing wholesale across the entire repo. What do you think @laurencejackson ?

Lu – can you upload some images of `img` generated using `rescale_to_byte` and your new version, to help illustrate how the data differs? I just want to get a better sense of what's changed.

hazenlib/ghosting.py
```diff
@@ -223,25 +244,28 @@ def get_ghosting(dcm, plotting=False) -> dict:
...
         except AttributeError as e:
             print(e)
             key = f"{dcm.SeriesDescription}_{dcm.SeriesNumber}"
         try:
-            fig, results[key] = get_ghosting(dcm, report_path)
-            if report_path:
-                fig.savefig(key + '.png')
+            fig, results[key] = get_ghosting(dcm, report_path=True)
+
+
```
tomaroberts, line 268: I don't follow the changes here? Possibly linked to my previous comment about the image `True.png` being output every time?

tests/test_ghosting.py
```diff
@@ -40,16 +44,18 @@ def test_calculate_ghost_intensity(self):
...
 
         assert 5.0 == hazen_ghosting.calculate_ghost_intensity(ghost=np.asarray([10]),
-                                                               phantom=np.asarray([100]),
-                                                               noise=np.asarray([5]))
+                                               phantom=np.asarray([100]),
+                                        noise=np.asarray([5]))
+
+            # assert -5.0 == hazen_ghosting.calculate_ghost_intensity(ghost=np.asarray([5]),
+            #                    phantom=np.asarray([100]),
+        #                   noise=np.asarray([10]))
```
tomaroberts: I think you can remove this – assume your abs() correction has made this redundant.

tests/test_ghosting.py
```diff
@@ -40,16 +44,18 @@ def test_calculate_ghost_intensity(self):
...
-                                                     phantom=np.asarray([-100]),
-                                                     noise=np.asarray([-5]))
+             phantom=np.asarray([-100]),
+         noise=np.asarray([-5]))
 
         assert 5.0 == hazen_ghosting.calculate_ghost_intensity(ghost=np.asarray([10]),
-                                                               phantom=np.asarray([100]),
-                                                               noise=np.asarray([5]))
+                                               phantom=np.asarray([100]),
+                                        noise=np.asarray([5]))
```
tomaroberts: Some of the indentation in this section is awry – would you mind just tidying.

tests/test_ghosting.py
```diff
@@ -79,44 +85,57 @@ def test_get_ghosting(self):
...
+
+        GHOSTING = (None,  0.007246960909896829)
+
+        def setUp(self):
+            self.file = str(TEST_DATA_DIR / 'ghosting' / 'GHOSTING' / 'axial_philips_broomfields.dcm')
+            self.dcm = pydicom.read_file(self.file)
+
+
+
+
```
tomaroberts: A lot of this indentation seems to be double-tabbed. Is that intentional? Github thinks the code has changed, but a lot of it looks the same from what I can tell.

tomaroberts (changes requested):
Thanks for all this Lu – see review comments. Main ones are that the code is always outputting `True.png` and it would be good to get a bit more clarity on the difference between `rescale_to_byte` and your version. Thanks!

hazenlib/ghosting.py
Lucrezia-Cester: Hi Tom, thank you for all the comments. No, this one is not intentional.

hazenlib/ghosting.py
Lucrezia-Cester: Hi Tom, the original rescale_to_byte function performs an histogram equalisation, which slightly manipulates the data. I think it is unnecessary to manipulate the data when we can just rescale in a more conventional way which does not manipulate the data. However, for this ghosting script, the rescale_to_byte is only used to plot, so actually I would leave the rescale_to_byte original function. I only changed this to check if it was making any difference while trying to fix the ghosting problem.

hazenlib/ghosting.py
Lucrezia-Cester: Previously I could not get the script to output an image, so that's why I changed it, but now I need to fix this problem of always outputting an image that I created. I'll fix this.

tests/test_ghosting.py
Lucrezia-Cester: yes I think so too. Will remove then.

tests/test_ghosting.py
Lucrezia-Cester: I'll tidy up the indentation.

tests/test_ghosting.py
Lucrezia-Cester: I'll tidy up the indentation.

PR #197, opened 2021-12-22
run cli tests as job in dev and production workflows

laurencejackson:
Small change to improve visiblity of tests in the actions tab.

Before, a separate CLI tests workflow was triggered on PRs or pushes to production branches. This was  duplicate of the triggers in theother workflows. This PR adds the CLI tests as a second 'job' performed in parallel with the unit testing.


Lucrezia-Cester (commented):
Hi Laurence, this looks good. I only have one question. Why is the operating system not including mac users? (Line 97)

laurencejackson:
> Hi Laurence, this looks good. I only have one question. Why is the operating system not including mac users? (Line 97)

I've just tried rerunning the tests with macos-latest included in the os-matrix but the macOS tests fail, see below:

image

This issue with the unit test jobs is that technically docker doesn't support macos. We get around this on our development machines by using the docker desktop app which runs containers in a Linux VM and is not something that can be installed through the github actions as it is not supported (I just learned this now!).

The issue for the unit test_cli job is that the psycopg2 package requires an SSL library that is not available by default on macOS. We get around this on our development laptops by installing OpenSSL through brew and linking these in the environment (there's a quick description of this in the readme). I don't think that currently, we want to spend the time installing brew and the SSL libraries to perform these tests as we expect them to be identical across unix OS's like ubuntu and macOS. 

Sorry for the long post. To summarise, I think testing with ubuntu only is sufficient for our purposes, and the added work of getting he tests to pass on macos is unlikely to offer us any benefit.

Lucrezia-Cester (approved):

laurencejackson:
Hi Lu, if you're happy with these changes please go ahead and merge them into the next release when you're ready. Thanks!

PR #255, opened 2022-07-22
syncs requirements with webapp

laurencejackson:


tomaroberts (approved):
Created new venv, built successfully and tests passing.

PR #257, opened 2022-07-25
Standardise output of slice position

sophie22:
return a nested dictionary (key:value pairs rather than a list)


laurencejackson:
resolved by #260

PR #261, opened 2022-07-26
Adds new docs

heyhaleema:
This PR adds:

- `newdocs` folder
- New `.rst` files with updated list of contributors, content from README.md and CONTRIBUTING.md, references, and task descriptions from the hackathon

Next step is to auto-generate `.rst` files for the `hazenlib` tasks but when I run `sphinx-apidoc -o newdocs/source/hazenlib hazenlib/`, it doesn't identify the functions and docstrings. Some of the errors I receive are:

image

image


newdocs/source/conf.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'hazen'
+copyright = '2022, Haris Shuaib'
+author = 'Haris Shuaib'
+
+# The full version, including alpha/beta/rc tags
+release = '0.6.0'
```
laurencejackson, line 25: We should try to make sure we always pull the version number from a single source, otherwise we can end up with the python package labelled having different labels.

I think the best place to get it from for now is setup.py in the root dir, you could try something like:

`from ...setup import __version__`

not sure i have the right number of dots or if a relative path import will work here, but you get the idea. let me know if that doesn't work and we can try something else!

newdocs/source/contributors.rst
```diff
@@ -0,0 +1,16 @@
...
+* `Tom Roberts <https://github.com/tomaroberts>`_
+* `Lucrezia Cester <https://github.com/Lucrezia-Cester>`_
+* `Paul Wilson <https://github.com/pcw24601>`_
+* `Yassine Azma <https://github.com/YassineRMH>`_
+* `Haleema Al Jazzaf <https://github.com/heyhaleema>`_
+* `Sophie Ratkai <https://github.com/sophie22>`_
+* `Ryan Satnarine <https://github.com/rs-sprout98>`_
+* `Jane Ansell <https://github.com/ansellj>`_
+* `Jamie Small <https://github.com/JamieSmall>`_
+* `Elizabeth Stamou <https://github.com/elizaGSTT>`_
```
laurencejackson, line 16: Can you also add some extra contributors:

Francesco Padormo <https://github.com/francescopadormo>
Sian Culley <https://github.com/superresolusian>

newdocs/source/gettingstarted.rst
```diff
@@ -0,0 +1,159 @@
...
+Clone the repository
+^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: bash
+
+   git clone https://github.com/GSTT-CSC/hazen.git
+
+Docker
+^^^^^^^^^^^^^^
+We recommend using the Docker version of *hazen* as it is easy to get up-and-running and is linked to the most stable release. Refer to the `Docker installation instructions <https://docs.docker.com/engine/install>`_ to install Docker on your host computer.
```
laurencejackson, line 28: we can leave this for now, but in future it'd be nice to be able to say we recommend the web app version of hazen!

laurencejackson (changes requested):
Thanks Haleema, sorry it took so long to get the review completed.

Comments attached

newdocs/source/gettingstarted.rst
heyhaleema: What do you think of adding a tip at the top of the page? For example:

image

@tomaroberts Is there a new Lightsail link we can use or should I hold off for now/use the Heroku link?

tomaroberts: We still need to decide on our long-term deployment strategy. I've shown POC that Lightsail will work. We need to work out a few details such as payment method and potential URL – Lightsail allows you to use a static IP, but would be preferable to point hazen.com or hazen.gstt.nhs.uk at it instead. Can obviously just use static IP in short-term if getting a URL is a challenge.

PS next task: I'm going to create the hazen-web-app staging/production pipelines, then it'll basically be good to go live in its current form.

PR #286, opened 2022-12-08
New spatial resolution Task

Lucrezia-Cester:
Porting the spatial_res function that I had made some time ago to the new Hazen


tomaroberts:
Lots of prior discussion on this PR in #141 

This implementation of a new spatial resolution task needs testing. The CLI would also need updating to give the option which spatial resolution test to use. Or, we choose one over the other. Unit tests also need to be created/updated.

tomaroberts:
Dormant PR, so closing for now.

PR #288, opened 2022-12-15
Adds .readthedocs.yaml to fix RTD build

heyhaleema:
This PR adds a ReadTheDocs (RTD) configuration file (see docs [here](https://docs.readthedocs.io/en/stable/config-file/v2.html) and closed issue [here](https://github.com/readthedocs/readthedocs.org/issues/6324)) since the build is currently failing due to outdated Python version being used in virtual environment (see error message [here](https://readthedocs.org/api/v2/build/18927072.txt)).


hshuaib90:
This branch is still failing to build on RTD
image

heyhaleema:
@hshuaib90 It looks like that build checked out `main` rather than `add-rtd-yaml`, so I did some testing on `fix-readthedocs` and it now works again there (see build [here](https://readthedocs.org/projects/hazen/builds/18937155/)). I think the main issue is that if you use the RTD configuration file, the default `sphinx` version is (assumedly) the latest, i.e. `5.3.0`. 

We used this configuration file in order to be able to install a later version of `scipy` with an updated Python version, but `sphinx 5.3.0` doesn't seem to support `sphinxcontrib-bibtex` (see build [here](https://readthedocs.org/api/v2/build/18936827.txt)). The fix was to include/uncomment the option to install `requirements.txt` so that we could pin the Sphinx version before the default was used/rest of the build continues.

hshuaib90:
I'm slightly confused. Why is the build on RTD not referencing this branch but the `fix_readthedocs` branch instead? How do we know this branch will build when it merges to main?

heyhaleema:
@hshuaib90 I think it's because we currently* need to explicitly activate the branches/versions we want RTD to trigger a build on:

image

We would ideally then activate the relevant development branch prior to merging with `main` (or `release` branch?) to reconfirm the RTD build hasn't broken, so if you activate `add-rtd-yaml`, we can see if it works there, too. The `latest` version hasn't yet passed successfully because of that issue with the newer`scipy` version which is why it's only passed on `fix-readthedocs` since that branch was using an older version.

\* It looks like we can automate this - see [here](https://docs.readthedocs.io/en/stable/automation-rules.html).

hshuaib90:
Yes you are right! Sorry, just built it and it works.

hshuaib90 (approved):

PR #292, opened 2022-12-29
Pre-release cleanup

heyhaleema:
This PR:
- Cleans up user instructions, i.e. removes `-v` from docstring ([docopt](http://docopt.org/) seems to only support `--version`?)
- Standardises task filenames and test directory structure for `acr_ghosting` and `acr_uniformity`
- Updates the `acr_ghosting` and `acr_uniformity` commands in the [CLI Test workflow](https://github.com/GSTT-CSC/hazen/blob/main/.github/workflows/test_cli.yml) 
- Updates version from `1.0.4` to `1.0.5` in `hazenlib/_version.py` and `CITATION.cff`
- Updates release instructions in `README`, i.e. files which import version number from `hazenlib/_version.py`
- Cleans up `.gitignore`
- Updates `hazenlib/__init__.py`*, `hazenlib/HazenTask.py`, `snr_map`, and other tests** so that all tasks and tests reports are saved to a `report` directory in the current working directory (i.e. `hazen/report/<task_name>`) and `hazen/tests/report/<task_name>`, respectively

\* I've added a comment in [this](https://github.com/GSTT-CSC/hazen/commit/2a1d9ec95a359fc3ddbc25c5be39e8ef898c7055) commit, but can/should we remove the `report_dir` arg in `hazenlib/__init__.py` (see line 366) since it's defined in HazenTask's init?
\** Similar to `snr_map`, do we want to refactor `relaxometry`, too, or leave as-is for now?


heyhaleema:
**Note:** CLI tests fails due to updates in filenames for `acr_ghosting` and `acr_uniformity` (see [this](https://github.com/GSTT-CSC/hazen/pull/292/commits/4aab2dff2192c154735be8b6ea06fdabf5c77ca5) and [this](https://github.com/GSTT-CSC/hazen/pull/292/commits/49829b453b9e1ff3185e67971e627b83cfb3f729) commit for more information). I've pushed [this](https://github.com/GSTT-CSC/hazen/pull/292/commits/81af26da2fe2fc523ce789ca766d38def0bbcbc6) and [this](https://github.com/GSTT-CSC/hazen/pull/292/commits/0b50eb18a19d126884b35fd896debe87406212db) commit to fix the issue.

tomaroberts:
Thanks for this @heyhaleema and for the lovely, clear PR bullet-points!

I'm reviewing now. One initial thing re: the data in the `hazenlib/data/relaxometry` folder. I've just checked some of the old Releases and that data has indeed always been in there, which is a surprise to me! Then looking in `hazenlib/relaxometry.py` it looks like the files are used as reference data for the relaxometry task. So let's leave it.

tomaroberts (commented):
Looks good, thanks. Two minor related comments:

---

**1)**

One thing I've noticed... there is a bit of inconsistency with the creation of the `report` folders. Currently, if I have a fresh local installation, then I have the following scenarios:

a) Run a hazen command **without** `--report option`, e.g. `hazen snr /path/to/data`

This creates an empty `report/SNR` folder.

b) Run a hazen command **with** `--report option`, e.g. `hazen snr /path/to/data --report`

This creates a `report/SNR` folder and saves data to the folder.

I'm not sure Option 1 is necessary? That said, it's not problematic really either.

---

**2)**

Slightly more odd – if I have a fresh install and then run `pytest tests/`, I get the following folders and files created:

Screenshot 2022-12-30 at 10 03 09

I think it would be better if it only creates the `tests/report` folder when we run pytest. For some reason it's creating a partial directory structure in `report`

---

If you could amend 2) that would be good Haleema, otherwise I'm not so bothered about 1). I think this shows the report generation in hazen could do with a bit of tightening, but it's not a major issue.

heyhaleema:
@tomaroberts I've pushed [this](https://github.com/GSTT-CSC/hazen/pull/292/commits/ca06714b84701c9553ba02bbda52aa9942194db0) commit, which seems to fix the issue, i.e. create the parent `report` and child task directories only when `--report` or `report_path=True` is specified in the command line or in the individual test files (e.g. `test_snr_map.py`), respectively.

tomaroberts (approved):
Great, thanks! Just a small note: `tests/report/SNRMap` still gets created when you run `pytest/tests`. This is fine though because the purpose of SNRMap is to produce these files.

PR #297, opened 2023-01-03
296 hazen incompatible with python 311

### LINKED ISSUE #296: hazen incompatible with python 3.11
**Describe the bug**
hazen incompatible with python 3.11  due to scikit image wheels not being updated yet .

The maintainers say release 0.20 is in the works and will be released soon. We will need to visit this when that release is out. Four now we will need to restrict hazen to  python <3.11

**To Reproduce**
Attempt to install scikit-image on pytohn 3.11

**Additional context**

laurencejackson:
Updates the packaging info to use pyproject.toml and setup.cfg which is the method preferred by setuptools.

Restricts Python version to <3.11


heyhaleema (changes requested):
Looks great 🔥 thanks for fixing this, @laurencejackson! Before we merge, can you please also update the version in [`CITATION.cff`](https://github.com/GSTT-CSC/hazen/blob/main/CITATION.cff) since this doesn't inherit from `_version.py`?

laurencejackson:
> Looks great 🔥 thanks for fixing this, @laurencejackson! Before we merge, can you please also update the version in [`CITATION.cff`](https://github.com/GSTT-CSC/hazen/blob/main/CITATION.cff) since this doesn't inherit from `_version.py`?

Yes! fixed now

heyhaleema (approved):

PR #299, opened 2023-01-03
ACR Slice Position Functionality

YassineAzma:
* Adds an ACR slice position script and associated unit tests that calculates the bar length difference (which is twice the slice position displacement) of slices 1 and 11 of the ACR phantom. This is done within the new HazenTask framework. The methodology is as follows:
   * Calculates the centroid of the object following morphological operations and a centre-of-mass calculation.
      * This step also produces a binary mask.
  * Uses line profiles and peak-finding operations to identify features in the phantom corresponding to the two wedges at the top of the phantom.
  * Draws interpolated line profiles (for subpixel measurement) through these wedges.
  * Uses circular shifting of these line profiles to find the pixel offset that produces the minimum difference between the two line profile.
  * Uses this pixel offset to calculate the bar length difference.
 
These results are visualised in the report, as shown below:
image
image


YassineAzma:
Note that there is a function outside of the class, find_n_peaks, which will be later refactored into a library of ACR specific functions, in the same vein as tools.py.

Lucrezia-Cester:
Hi @YassineRMH,

I am having trouble checking this from the images, but you are basically calculating difference in length of the two wedges right? Such as shown in picture below

If that is the case, could this be incorporated as a "short summary" of the algorithm in the docstring?

Screenshot 2023-01-24 at 11 31 17

YassineAzma:
> Hi @YassineRMH,
> 
> I am having trouble checking this from the images, but you are basically calculating difference in legth of the two wedges right? Such as shown in picture below
> 
> Screenshot 2023-01-24 at 11 31 17

Pretty much. Measurement of wedge lengths didn't perform so well in low SNR datasets, and this method allows sub-pixel measurement via interpolation (similar to how it would be done in a DICOM viewer). I'll add axis labels to the plots with the x units being relative pixel position in mm. Would that make it easier to understand?

Lucrezia-Cester:
@YassineRMH  But the final number for slice position is the difference in length of the wedges right?

YassineAzma:
> @YassineRMH But the final number for slice position is the difference in length of the wedges right?

Yes, but I calculate it by finding the minimum difference in the subtraction of the right line profile through the wedge (subject to various shifts) and the left profile through the wedge.

Here's the updated docstring:

``` 
This script calculates the bar length difference in accordance with the ACR Guidance. Line profiles are drawn 
vertically through the left and right wedges. Circular shifts are applied to the right wedge's line profile and 
subtracted from the static left wedge line profile. The shift used to produce the minimum difference between the 
circularly shifted right line profile and the static left one are used to determine the bar length difference, 
which is twice the slice position displacement. The results are also visualised.

Lucrezia-Cester:
Hi @YassineRMH, could you please expand on this sentence as I cannot quite understand it "Circular shifts are applied to the right wedge's line profile and
subtracted from the static left wedge line profile."

I will then merge this PR.

YassineAzma:
> Hi @YassineRMH, could you please expand on this sentence as I cannot quite understand it "Circular shifts are applied to the right wedge's line profile and subtracted from the static left wedge line profile."
> 
> I will then merge this PR.

How about this:

This script calculates the bar length difference in accordance with the ACR Guidance. Line profiles are drawn 
vertically through the left and right wedges. The right wedge's line profile is shifted and wrapped round before being subtracted from the left wedge's line profile, e.g.:

Right line profile: [1, 2, 3, 4, 5]
Right line profile wrapped round by 1: [2, 3, 4, 5, 1]

This wrapping process, from hereon referred to as circular shifting, is then used for subtractions.

The shift used to produce the minimum difference between the circularly shifted right line profile and the static left one are used to determine the bar length difference, which is twice the slice position displacement. The results are also visualised. 

If this is acceptable, I'll commit it. Let me know where it does not make sense.

Lucrezia-Cester:
@YassineRMH thank you @YassineRMH, please commit this and I'll merge

Lucrezia-Cester (approved):

PR #305, opened 2023-01-10
Refactor hazen task to snr map

### LINKED ISSUE #302: Refactor snr_map to use HazenTask class
Refactor snr_map to use HazenTask class

pcw24601:


Lucrezia-Cester:
Hi @pcw24601, could the ghosting image be removed from this PR please?

pcw24601:
> Hi @pcw24601, could the ghosting image be removed from this PR please?

I've done that--not sure how it ended up in the commit, well spotted.

hazenlib/tasks/snr_map.py
```diff
@@ -1,11 +1,352 @@
...
+    ----------
+    vals : array-like
+        Array-like object to calculate sample stdev.
+
+    Returns
+    -------
+    standard deviation : ndarray, same dtype as vals.
+    """
+    return np.std(vals, ddof=1)
+
```
Lucrezia-Cester, line 45: Why is this function outside the class?

pcw24601: The function is used as an argument to `scipy.ndimage.filters.generic_filter` on L127, so it can't be a method (as the argument list would include `self`).

I can try changing it to a `@staticmethod` within the class next week.

Lucrezia-Cester: ok, shall I then wait before merging?

pcw24601: I'm happy either way. It should only take a few seconds to change (and a few minutes to test), but I need to do it from my home machine. I think it's a design decision that I don't have enough experience to know which way is 'best' and will follow your preference.

I should be able to correct it before the end of Tuesday at the latest. Let me know what you think.

Lucrezia-Cester: thanks @pcw24601 for letting me know. We are planning a new release on Thursday so I will keep this PR open until then. If you fix it on Tuesday then I'll merge it then and it will contain the fix, if not I will merge it Thursday as it is in the new release and your edit will be added to the next release at the end of February.

pcw24601: I've removed the standard deviation function from outside the class. It was such a small function, and only used once, so I replaced it with a lambda function.

laurencejackson (approved):
Tested and looks good to me, thanks @pcw24601

PR #317, opened 2023-01-31
316 cli parser tests fail in windows path error

pcw24601:
Fixes platform specific paths in test that cause errors in Windows.


pcw24601:
Ready for review

heyhaleema (commented):
Hi @pcw24601, I've had a look at the changes and I'm just wondering if this will cause issues with non-Windows operating systems? If so, I think we should use something like [`platform.system()`](https://docs.python.org/3/library/platform.html#platform.system) to concatenate either backward or forward slashes depending on the OS detected.

pcw24601:
Hi @heyhaleema, thanks for the review. I don't think it's a problem as (1) it passes the GutHub development tests for ubuntu 3.9 and 3.10 and (2) the paths now consistently use the device-independent `pathlib.Path` objects which should take care of the platform differences (see definitions in `hazen/tests/__init__.py`). The use of `str(DIR / 'subdir')` caught me out (someone else had introduced this notation in some places in the codebase)--here the `/` operator is overloaded in `pathlib.Path` to join the paths in a device-independent fashion.

Lucrezia-Cester (commented):
this makes the path platfrom-independent, tests pass so i approve the pr

Lucrezia-Cester:
i also changed base

Lucrezia-Cester:
this branch was also out of date with main so I merged the new release changes to avoid merge conflicts

Lucrezia-Cester (approved):

PR #321, opened 2023-02-20
ACR Slice Thickness Functionality

YassineAzma:
Adds an ACR Slice Thickness script and associated unit tests that calculates the slice thickness of the ACR phantom for slice 1. This is done within the new HazenTask framework. The methodology is as follows:
   * Calculates the centroid of the object following morphological operations and a centre-of-mass calculation.
      * This step also produces a binary mask.
  * Uses line profiles to find the location of the ramps within the phantom.
  * Draw line profiles along the length of the ramps within the phantom using different offsets around the ramp locations found in the previous step.
  * The FWHM is determined for each line profile.
  * Calculates the slice thickness, dz, using the following formula for all line profile configurations:
$$dz = 0.2 * (L_{upper} * L_{lower})/(L_{upper} + L_{lower}) $$
  * Select the closest dz to the true slice thickness, as defined in the DICOM metadata, as the final result.
 
The results are visualised in the report. Examples are shown below:
image
image


YassineAzma:
Hi, @Lucrezia-Cester. This PR is ready for review whenever you're available.

Lucrezia-Cester (approved):

PR #323, opened 2023-02-27
Increase starting threshold to 0.5 quantile

### LINKED ISSUE #322: Li threhsold too low
**Describe the bug**
The current Li threshold starting estimate is the 0.25 percentile of pixel values, this can cause an error when an image with lots of zeros is thresholded as the staring value can equal 0.0. 

**To Reproduce**
Run slice position test on an image series where the 0.25 percentile in any slice is 0.0.

**Expected behavior**
The threshold should be set higher so that the Li threshold optimisation can start.

**Screenshots**
image

**Desktop (please complete the following information):**
Bug present in hazen 1.1.2

laurencejackson:
Resolves #322 by increasing Li threshold starting value to 0.5 quantile.


tomaroberts (approved):
LGTM

PR #331, opened 2023-04-12
updated version

Lucrezia-Cester:
Lat time I made a release I updated the version number only after I had merged into main and only after I had created a new release and this caused issues. 

I am now adding a PR to specifically change the version so that it is updated in PiPy.

@laurencejackson, @tomaroberts, could you check this works ?


tomaroberts (changes requested):
These changes are in the right places, but the notation should be "1.2.0" rather than "1.2" for consistency with the other versions. Can you change and then it can be merged.

Lucrezia-Cester:
@tomaroberts, Hi Tom, I have updated the version nomenclature. Can I now do a new release? Is this all that is required to ensure it updates PyPi as well?

tomaroberts:
Yep – that should be sufficient for release now.

Issue #332, opened 2023-04-13
acr geometric accuracy module 


elizaGSTT:
as seen in the screenshot there is bad centering and results are therefore far off. @YassineAzma 

<img width="617" alt="Screenshot 2023-04-06 at 10 45 02" src="https://user-images.githubusercontent.com/93722461/231781745-3edd17b4-4ef2-4d64-bf74-caf00b52328b.png">

YassineAzma:
Resolved in #343.

Please don't acquire with just the spine coil.

PR #343, opened 2023-05-11
ACR v1.1 - Refactor and Improved Masking

### LINKED ISSUE #333: acr ghosting module
problem centering ROIs. this datset is from bodycoil acquisition. @YassineAzma 
Screenshot 2023-04-06 at 10 40 41

YassineAzma:
This PR introduces the ACRTools class in acr_tools.py, which contains a number of useful functions that are frequently used throughout the ACR analysis scripts.

**sort_images** = DICOM and image sorting via the image position

**determine_rotation** = Estimate the rotation angle of the phantom using the Hough circle detector

**rotate_images** = Rotates the images by the rotation angle

**mask_image** = Thresholds the image - if the intensity variation in the image is sufficiently large, local thresholding methods are used rather than a global intensity threshold.

**find_phantom_center** = Now decoupled from the masking routine. The image is blurred and the edges are then sharpened along the diagonal. The Hough circle detector is then used to identify the phantom using prior knowledge of its geometry.

**circular_mask** = Draw a circular mask

**measure_orthogonal_lengths** = Measure the horizontal and vertical extent of the mask.

**rotate_point** = Find x' and y' for a point rotated around a given centre by a given angle.

**find_n_highest_peaks** = Given an array, report the location and intensity of the highest peaks.

Along with other general improvements, these functions have been used to refactor the existing ACR functions. This will prevent unneeded repetition in the unit tests for the individual analysis scripts by grouping tests for reused functions in the acr_tools.py module.

**NOTE**: All ACR test outputs have changed by <2% owing to changes in the masking and centre definition methods. These changes should make them more robust to issues seen in Issues #333 and #332.

Pending: Complete unit tests for acr_tools module.

Resolves #333, #332


YassineAzma:
@Lucrezia-Cester I can't seem to find where the test_tools.py unit tests are called in the workflow tests. Could you please point me in the right direction to do something for my test_acr_tools?

tomaroberts (changes requested):
@YassineAzma – this is an amazing PR frankly. Crazy amount of work gone into this. Apologies it's taken us so long to review. Thank you so much for doing it! 🎉🎉🎉 

I've run the unit tests locally and they are all passing.

I've also run `hazen acr_XXX` one-by-one on each of the ACR Tasks. They all work on the Siemens test dataset you've provided and they all provide output report files.

A couple of minor things I've spotted:
- There is a 'rotation angle' message printed during the `acr_spatial_resolution` task. I think there is an error with the wording which needs fixing. See here: image
- During the `acr_snr` task, there is a matplotlib related message printed. See below. Screenshot 2023-07-17 at 16 33 50

EDIT: For both of the above Yassine, can we have the warning output to the logger? It's useful information for the user, but would be good to keep the output clean. Another thought as well: if the rotation angle is genuinely useful to know, could it be passed as an output of the Task?

- In the `test_geometric_accurary` and `test_acr_uniformity` files, there are commented lines referring to Philips unit tests. Is this a reminder for the future? If so, perhaps just adjust to "TODO" which is a special type of comment for this kind of purpose.

Again, this amazing. Thank you!

tomaroberts:
Just to add to this, you've written a fair amount of documentation in the files, which is great. When we come to fully uplifting the hazen ReadTheDocs, we'll definitely need your input on making the guidance as clear as possible. Thanks.

tomaroberts (approved):

hazenlib/tasks/acr_geometric_accuracy.py
```diff
@@ -169,72 +112,92 @@ def diagonal_lengths(self, res, mask, cxy):
 
     def get_geometric_accuracy_slice1(self, dcm):
         img = dcm.pixel_array
-        res = dcm.PixelSpacing
-        mask, cxy = self.centroid_com(img)
 
-        h_dict = self.horizontal_length(res, mask, cxy)
-        v_dict = self.vertical_length(res, mask, cxy)
+        mask = self.ACR_obj.mask_image(self.ACR_obj.images[6])
```
sophie22, line 116: Is it meant to be images[6] or the pixel array of the DICOM input to this function?

PR #351, opened 2023-07-06
350 utils

### LINKED ISSUE #347: Improve/fix usage instructions - separate init.py and main.py
Screenshot 2023-06-30 at 10 57 15

there are typos in the current instructions and could be arranged in a more logical order

sophie22:
## Summary
in line with Python best practice, remove all commands possible from the __init__.py into a new __main__.py script, make required changes to entry point and subsequently update the READme with the relevant commands
EDIT: upon further reading of documentation, looks like there is no real benefit to having the main function in a separate script from init.

combine utility functions from __init__, tools and shapes.py into a new utils.py, make required changes to all task scripts that use utility functions and their applicable test scripts

resolves #347 

Outstanding TODO:
- update original test_tools and test_hazenlib scripts and correctly group unit tests in line with new utils.py and main.py


sophie22:
https://realpython.com/python-main-function/
main function is not supposed to have a return value, but could call other functions that do --> will have to update the test_hazenlib unit tests that rely on evaluating the stdout to check output is as expected

tomaroberts (commented):
Changes look great.

Just need to resolve the [GitHub Actions issue where the CLI tests fail](https://github.com/GSTT-CSC/hazen/actions/runs/5528992901/jobs/10086452799?pr=351#step:6:22).

sophie22:
as discussed last week, GitHub tests fail either the CLI test when tasks complete with exit code 1 due to a string being returned from the `main` function, or the development tests fail: 2 tests within test_utils rely on having the string returned by main to be compared against expected values.

So to resolve these, I plan to change the `main` function so it doesn't return anything and printing of the result is handled before the end of `main` and then modifying the test_hazenlib functions to parse sys.stdout to capture the result printed.
https://pytest.org/en/7.4.x/how-to/capture-stdout-stderr.html

@tomaroberts let me know if this sounds sensible and I'll get on it

tomaroberts:
As discussed on Teams, give it a go :)

sophie22:
made the agreed modifications such that hazenlib.main() does not return anything (completes with exit code 0)
and modified the tests for the SNR and relaxometry tasks such that they do not rely on hazenlib.main() which now doesn't return anything, but instead call their respective run() function with equivalent arguments as passed to CLI parser.

sophie22:
in the future, I recommend to separate hazenlib.main() into the following sub-functions:
* `parse_CLI_args`: purely responsible for presenting a useful `help` and parsing the passed arguments
    - it allows for extendability by adding new tasks or new optional arguments to existing tasks easily
* `prepare`: all sorts of checks on the inputs provided: includes checking that an available task was requested with applicable optional arguments, and that suitable DICOMs exist within the folder specified. Prepare dicom files to be passed to a task and create report output folder, if applicable.
* `run_task`: make an instance of the selected task class and call its run() function with arguments as identified above, this is what would return the `result` dictionary that is printed within the main() function, which returns nothing
    - this is what could be tested in the unit tests that task with known input returns calculated values asa expected

tests/test_hazenlib.py
```diff
@@ -46,26 +43,28 @@ def test_main_snr_exception(self):
...
 
     def test_relaxometry(self):
         path = str(TEST_DATA_DIR / 'relaxometry' / 'T1' / 'site3_ge' / 'plate4')
-        sys.argv = ["hazen", "relaxometry", path, "--plate_number", "4", "--calc_t1"]
-
-        output = hazenlib.main()
-        output_dict = ast.literal_eval(output)
+        files = get_dicom_files(path)
+        dicom_objects = [pydicom.read_file(x, force=True) for x in files if is_dicom_file(x)]
+        # sys.argv = ["hazen", "relaxometry", path, "--plate_number", "4", "--calc_t1"]
```
tomaroberts: Think we can remove this comment?

tests/test_hazenlib.py
```diff
@@ -46,26 +43,28 @@ def test_main_snr_exception(self):
 
     def test_snr_measured_slice_width(self):
         path = str(TEST_DATA_DIR / 'snr' / 'GE')
-        sys.argv = ["hazen", "snr", path, "--measured_slice_width", "1"]
+        files = get_dicom_files(path)
+        # sys.argv = ["hazen", "snr", path, "--measured_slice_width", "1"]
```
tomaroberts: Think we can remove this comment?

tests/test_hazenlib.py
sophie22: sure, wasn't sure if reference of the original was required to be kept. also left it to check that I made an equivalent call.
happy to remove these comments

tomaroberts (changes requested):
Great! :)

Couple of lines to remove, unless you think useful to keep.

Go ahead and merge `main` into this branch and resolve any conflicts.

tomaroberts (approved):

PR #354, opened 2023-07-14
156 only add "report" section to results when requested

sophie22:
tiny change as discussed in the Hazen meeting to standardise when and how the `"reports"` section with file path(s) to report images are added to `the results` dictionary in all tasks.
remove the creation of a report folder from the snr_map task as it is being created upfront within `__init__.py`


hazenlib/tasks/snr_map.py
tomaroberts, line 1: @sophie22 – I notice you removed the `if self.report` entirely from this file. Is this intentional? Is there no report associated with the snr_map task?

tomaroberts (commented):
Nice and easy to review – thanks for separating out :)

Just a small query regarding one file, otherwise I'll merge.

sophie22:
regarding the snr_map script: I only removed the report folder creation section, as it is being handled centrally in init.py. Further down around line 314 there still is the section for if.report to generate the diagrams/images, intact.

tomaroberts (approved):

Issue #360, opened 2023-07-24
Report images for relaxometry task


sophie22:
Currently the relaxometry task is not yet refactored to inherit from the HazenTask class and it's figures and diagrams generated with the `--show_rois`, `--show_template_fit` and `--show_relax_fits` options are not available. Currently matplotlib is not configured to be able to make pop out windows to show these diagrams and it's not implemented to save these figures.

I propose to refactor the report generation part of the relaxometry task to follow the procedure of other tasks and save the figures to file with the `--report` option, replacing the current the individual options.

tomaroberts:
Resolved by #361

PR #361, opened 2023-07-24
360 relaxometry task report

sophie22:
Potential fix for #360 by partially refactoring the relaxometry task to be closer aligned with other tasks, and especially fixing the way report images are generated and saved, rather than attempted to be displayed.


sophie22:
various tests are failing as expected. Will update the hazenlib test so that CLI args align with what is now expected to be provided for the task and will look into the relaxometry unit tests too.

sophie22:
I am unable to run the T2 calculation on my laptop due to a Bus error, it is not expected to have been introduced with these code changes. - will investigate further

tomaroberts (approved):
Tested locally and working. Tests passing. Thanks for changes.

PR #362, opened 2023-07-25
360 refactor relaxometry

sophie22:
part 1: simplify existing functions and make them more explicit


sophie22:
Part 1: simplification and adjusting unit tests is COMPLETE

TODO: consider adding back `time_attribute` to be an object attribute so it doesn't have to be passed each time, especially in the unit tests.

tomaroberts (commented):
Looks good – I've made a minor change to the filename. "config" is quite a loaded programming term in repos, normally indicates a file responsible for controlling parameters that you might want to adjust/not hard-code. Given the `config.py` was just predominantly parameters for the relaxometry Task, I've just updated the filename to reflect that.

tomaroberts (approved):

sophie22:
I was planning to move other parameters like ROI sizes and default/template values used in other tasks to the config file, but can still rename this file later if the move does happen.

sophie22:
I have 4 more commits that ended up not being pushed, I apologise. Will open a new PR

Issue #378, opened 2023-09-15
ACR geometric accuracy potentially using the wrong slices


sophie22:
**Describe the bug**
Based on the documentation and earlier version of the geometric accuracy script, the `get_geometric_accuracy_slice1` and `get_geometric_accuracy_slice5` functions should create a mask and find the centre based on slice 1 and 5 respectively. However, in the current implementation (since the refactoring) the mask and centre is created for slice 7, and distances are measured on slice 7 rather than 1 and 5.

Previous code: https://github.com/GSTT-CSC/hazen/blob/3ee1bb6b82385daf79d4b786a9660c8d805e11ad/hazenlib/tasks/acr_geometric_accuracy.py#L170
Current code: https://github.com/GSTT-CSC/hazen/blob/main/hazenlib/tasks/acr_geometric_accuracy.py#L116

**Expected behaviour**
Please confirm whether mask and length calculation should be performed on slice 7 and overlayed on slice 1 and 5 for the report images, or calculated separately for slices 1 and 5.

**Additional context**

YassineAzma:
This is a bug. It will only have impact on the results if there is any form of tilt between slices due to a positioning error.

sophie22:
Thank you for your prompt reply. Is it simply a matter of replacing slice 7 with 1 and 5 to create the mask from?

YassineAzma:
> Thank you for your prompt reply. Is it simply a matter of replacing slice 7 with 1 and 5 to create the mask from?

Yes, I think I ended up defaulting to slice 7 by mistake when refactoring because I had the idea that a flat image would be far easier to mask than the busier slices 1 and 5.

sophie22:
Is the centre expected to be the same across all slices?
I'm getting slightly different values especially for the vertical distance in the unit tests. Is there an alternative method for calculating the 'true' values to set in the unit tests?

YassineAzma:
> Is the centre expected to be the same across all slices? I'm getting slightly different values especially for the vertical distance in the unit tests. Is there an alternative method for calculating the 'true' values to set in the unit tests?

How much is the centre changing? 1 or 2 pixels in either direction?
How much are the length values changing by? I wouldn't be too concerned if it's less than 1mm and remains within tolerance.

[...]

PR #399, opened 2024-01-04
378 geom accuracy wrong slices

mollybuckley:
The ACR geometric accuracy test was measuring from slice 7 rather than slices 1 and 5 as per the ACR guidance. I have:

- fixed this, so that the measurements were coming from the correct slices
- combined the separate functions for ACR geometric accuracy slice 1 and 5 (the previous difference was that the slice 5 function was taking diagonal measurements too but the slice 1 function was not). The function now has slice index as an input and if the slice index is 4 then the diagonal measurements are made.

I then realised that the mask was being created incorrectly on slice 1 of the GE data (this was not previously an issue as the slice was being created from slice 7). This was because one of the pixel values outside the phantom was above the masking threshold value. I have changed the threshold value so that it is no longer picking up background noise pixels and including these in the mask. The images on the report look acceptable but some future work could be to test this with more GE data. I then changed three of the unit tests accordingly, as there were very minor variations in the results following this change.


sophie22:
resolves #378

hazenlib/ACRObject.py
```diff
@@ -124,7 +124,7 @@ def rotate_images(self):
 
         return skimage.transform.rotate(self.images, self.rot_angle, resize=False, preserve_range=True)
 
-    def find_phantom_center(self):
+    def find_phantom_center(self, img):
```
sophie22, line 147: this is great, will have to update the docstring for the function

hazenlib/ACRObject.py
```diff
@@ -151,7 +150,7 @@ def find_phantom_center(self):
         radius = int(detected_circles[2])
         return centre, radius
 
-    def get_mask_image(self, image, mag_threshold=0.05, open_threshold=500):
+    def get_mask_image(self, image, mag_threshold=0.07, open_threshold=500):
```
sophie22, line 175: it would be great to be able to test what this threshold value should be on a range of datasets from various scanners with various magnetic strengths. - not a blocker for merging this PR but good for fine-tuning the measurements

hazenlib/ACRObject.py
```diff
@@ -203,7 +202,7 @@ def circular_mask(centre, radius, dims):
 
         return mask
 
-    def measure_orthogonal_lengths(self, mask):
+    def measure_orthogonal_lengths(self, mask, slice_index):
```
sophie22, line 241: you could provide the img (pixel_array) as input directly, not just its index

hazenlib/ACRObject.py
```diff
@@ -227,16 +226,19 @@ def measure_orthogonal_lengths(self, mask):
         """
         dims = mask.shape
         dx, dy = self.pixel_spacing
+        [centre, radius] = self.find_phantom_center(self.images[slice_index])
```
sophie22: I think these can be unpacked directly in one line, like
`[(vertical, horizontal), radius] = self.find_phantom_centre(...)`

hazenlib/tasks/acr_geometric_accuracy.py
```diff
@@ -85,97 +84,92 @@ def run(self) -> dict:
...
-            fig.savefig(img_path)
-            self.report_files.append(img_path)
-
-        return length_dict['Horizontal Distance'], length_dict['Vertical Distance']
 
-    def get_geometric_accuracy_slice5(self, dcm):
-        img = dcm.pixel_array
-        mask = self.ACR_obj.get_mask_image(self.ACR_obj.images[6])
-        cxy = self.ACR_obj.centre
+    def get_geometric_accuracy(self, slice_index):
```
sophie22, line 100: this is amazing and very useful for future code maintenance that you combined these functions!

sophie22 (changes requested):

hazenlib/ACRObject.py
mollybuckley: done

tests/test_acr_ghosting.py
```diff
@@ -33,7 +33,7 @@ def test_ghosting(self):
 
 class TestACRGhostingGE(TestACRGhostingSiemens):
     centre = [253, 256]
-    psg = 0.471
+    psg = 0.482
```
sophie22, line 41: please explain why these values had to be updated?

mollybuckley: The ACR ghosting and ACR slice position unit tests were both updated because they both utilise the `get_mask_image` function in the ACR object, where the threshold was changed from 0.05 to 0.07 in this PR. This lead to very minor changes to the numerical results of these other tasks.

sophie22 (approved):
Thank you! All comments have been satisfactorily addressed.
Happy to merge :)

PR #409, opened 2024-01-29
ACR slice ordering

### LINKED ISSUE #404: ACR object slice ordering results in use of incorrect slice
from [RebeccaThornley](https://github.com/RebeccaThornley) commented [on Dec 7, 2023](https://github.com/GSTT-CSC/hazen/pull/353#issuecomment-1844918063)

Hi, on a recent data set the slice order inversion did not seem to work for some tests: SNR, ghosting, uniformity. 

Though inversion seem to be applied correctly for geometric accuracy, the measurements were not performed correctly/not displayed correctly on the report. 

ACR_SAG_Rotate_90_new_1017001_7
ACR_SAG_Rotate_90_new_1017001_11

_Originally posted by @RebeccaThornley in https://github.com/GSTT-CSC/hazen/issues/353#issuecomment-1844918063_

sophie22:
This PR includes important changes to how the ACR phantom image slices are being processed prior to specific task measurements.

### An instance of the ACRObject class now has the following attributes:
- `dx` and `dy` corresponding to the PixelSpacing values from the DICOM header (in x and y directions)
- `slice_stack` is a list of the input DICOMs, ordered by position (from ImageOrientationPatient) and ordered by phantom orientation (see below)
- there is no LR flipping of pixel values

### The following methods are available for ACRobjects:
- `sort_dcms`: orders DICOMs based on the changing coordinate according to acquisition orientation
- `order_phantom_slices`: orders DICOMs based on phantom orientation - first slice is expected to have a circle while the last one is expected not to have one
- `find_phantom_centre`: locate the coordinates of the phantom (circle) in the image and also return its radius
- `get_mask_image`: pixel array with values above given threshold from input image
- `circular_mask`: list of coordinate matrices from coordinate vectors with given centre and radius
- `determine_rotation`: determines rotation angle of phantom images
- `rotate_images`: rotate the pixel array of all phantom slices by a defined angle
- `rotate_point': calculate new coordinates of a point (pixel) after rotation by a defined angle
- `measure_orthogonal_lengths` and `find_n_highest_peaks` are specific to task measurements

## Other important  changes:
In task-specific classes, the phantom centre is now being calculated based on the specific slice, not taken from the flood field slice as previously.
Resolution is now being used from the ACRObject dx, dy attributes rather than being accessed from the DICOM header values.
`determine_orientation` and `detect_circle` are new functions moved to the utilities to make them more widely reusable.
`scipy` version was updated to make use of a predefined function rather than specifying our own.
Unit tests were simplified for ACR tasks and new tests were added for ACRObject for a more comprehensive coverage of this class.
Some test values had to be updated to account for different rounding methods, but all are within tolerance as per ACR guidelines.


sophie22:
- nothing works for Sagittal due to a 90 degree anticlockwise rotation.
- all tasks (and especially their report images) for all orientation need to be checked - things like correctly shifting the centre coordinates, sampling from relevant regions, etc
- report image generation should be split and moved to their respective functions calculating the values being represented

sophie22:
all tests are failing as these haven't been updated to use the new functions yet

sophie22:
Need to update a couple of measurement values - will be checked by @mollybuckley on 14 Feb, 2024.

Screenshot 2024-02-13 at 16 43 25

sophie22:
test_ACRObject is missing the test datasets currently

sophie22:
expected values were updated following discussion with Molly, having concluded that differences are due to rounding and are well within the tolerances for their respective measurement as per ACR guidelines.

hazenlib/ACRObject.py
```diff
@@ -208,7 +206,7 @@ def get_mask_image(self, image, mag_threshold=0.07, open_threshold=500):
 
     @staticmethod
     def circular_mask(centre, radius, dims):
-        """Sort a stack of images based on slice position.
+        """__meow__.
```
tomaroberts: meow? Should this be in here?

hazenlib/utils.py
```diff
@@ -328,6 +329,69 @@ def get_image_orientation(iop):
         return "Transverse"
 
 
+def determine_orientation(dcm_list):
```
tomaroberts, line 332: This is not checking the phantom orientation because that depends on how the user put the phantom in the scanner. The `ImageOrientationPatient` tag tells you the orientation of the _imaging volume_ which contains the phantom. If the 'transverse' orientation of the phantom is aligned with the transverse definition of the imaging volume (e.g. [1, 0, 0, 0, 1, 0] in the comments below) then it couples that the phantom is in the transverse orientation.

However, the user could place the phantom at 90 degrees to 'transverse', but the same imaging volume might be applied. In this case, [1, 0, 0, 0, 1, 0] still reflects a transverse imaging volume, but this does not match the phantom. You have no way of knowing the **true** orientation of the phantom from the DICOM metadata.

There is also an added complication in all of this because different scanner vendors define there scanners axes in different ways. The FH (Foot-Head) direction on a Philips scanner might be defined as HF (Head-Foot) on a Siemens scanner.

My feeling is that if the unit tests are passing with the code below, then keep it in, assuming it doesn't affect anything else in hazen. But be mindful that this could be a source of bugs in the future. Which, might have also been present with the previous implementation.

tomaroberts (commented):
@sophie22 – I think this is fine. Bit tricky to tell the extent of the changes, but if the tests are passing, then that's okay.

One comment about the phantom_orientation function. Ultimately, the user should configure the phantom in an orientation compliant with the guidelines, but if this helps iron out some issues, then go ahead.

hazenlib/ACRObject.py
sophie22: @mollybuckley please help to provide a better description of this function

hazenlib/utils.py
sophie22: added this caveat to the docstring so it will be apparent if causing issues in the future.
will also include in the Release notes that it was only tested on Transverse images from Siemens and GE scanners.

tomaroberts (approved):

hazenlib/utils.py
sophie22: some of this caveat did become an apparent problem and is fixed in #416

Issue #352, opened 2023-07-13
ACR - Sagittal and Coronal Planes 


YassineAzma:
**Is your feature request related to a problem? Please describe.**
Some sites like to use the ACR phantom in additional planes to fully characterise the gradient system. This is not recommended in the guidance, which specifies that acquisitions should be performed in the axial plane, but would be a nice and relatively easy feature to implement in Hazen.

**Describe the solution you'd like**
For coronal acquisitions, transform the images into the standard view.

For sagittal acquisitions, transform the images into the standard view. This will likely require a rotation.

All operations will then be subject to a slice ordering and LR orientation check, which is being implemented to resolve #345.
 
**Describe alternatives you've considered**
There is no alternative bar requiring users to manually adjust images prior to running through Hazen. This is not ideal for larger datasets.

**Additional context**
Fully explained above.

RebeccaThornley:
To add to this from recent QA experience- it would be really useful to incorporate a rotation into the code where needed. i.e. to either recognise a wrong orientation and apply rotation automatically or to give as an optional input to the user. For the sagittal acquisition, the images require a CW 90degree rotation in order to be in the correct orientation for analysis with hazen. Whilst on Siemens scanners we can manually do this rotation at the scanner before export, we have not yet found a way to do this with Philips (or other manufacturers). See example data set acquired sagitally with a Philips scanner. 
[SAG_CLEAR_1.zip](https://github.com/GSTT-CSC/hazen/files/14293698/SAG_CLEAR_1.zip)

sophie22:
Hi Becky, thanks for the feedback and providing data from QA scenarios. Could you please share additional datasets by email of ACR phantom images acquired in all orientations and if possible with and without manual rotation applied?
I'm currently looking into improving this part of the hazen/ACRObject functions, would be great to have a range of data to test with!

sophie22:
Thank you for sending the datasets by email.
COR, TRA and the manually rotated SAG all work well as expected (for slice position task) following the changes I made in the related branch.
Will look further into correctly rotating the original SAG dataset.

There are additional issues with the TRA and COR datasets with the slice thickness task, where indexes on the generated line profile to calculate FWHM is out of range - I'm investigating now.

PR #416, opened 2024-02-29
ACR Object orientation - bugfix

sophie22:
Account for fluctuation in ImagePositionPatient values in all directions - assume that "true" orientation is where values across the image set are changing.
When the orientation cannot be determined from DICOM metadata values, error out and do not attempt to assume orientation. - In the future, provide an optional CLI argument to override orientation and provide the axis to be used for orientation.


mollybuckley (approved):

sophie22:
resolves #352

PR #417, opened 2024-02-29
ACR slice thickness improvement

sophie22:
Check if expected x_pts (indices) will be out of range, ie greater than len(ydata) which is the line profile.


mollybuckley (approved):

sophie22:
resolves #352

PR #422, opened 2024-03-05
Address python version compatibility

### LINKED ISSUE #412: Incompatible python versions
**Description**
A few members of our team have freshly downloaded hazen recently using versions of python higher than 3.9. Using the pip install with these later versions of python results in an old version of hazen being downloaded (1.0.2). Both 3.11 and 3.12 have been tested. 

**To Reproduce**
Following pip install guidance:

python3 -m venv hazen-venv [or python3.11(12) -m venv hazen-venv]
source hazen-venv/bin/activate
pip install hazen

**Expected behavior**
Currently instructions say that python version of 3.9 or higher is required. Therefore expected v3.11 or v.12 to be compatible. 

**Screenshots**
Screenshot 2024-02-15 at 09 50 22

**Desktop:**
 - OS: macOS SOnoma v14.2.1
 - Python version: 3.12
 - Hazen version: results in download of 1.0.2
 - Deployment: venv

**Additional context**
This isn't a big problem as the easy work around is to download python3.9 and install hazen with this. But if it is the case that 3.9 is a requirement could the instructions be updated to be more clear?

sophie22:
Extend PyPI support up to Python 3.11.x
Tested on Mac with Python 3.11.5 and on Windows via Anaconda with Python 3.11.7
3.12 has incompatible built-in packages and is therefore not currently supported.


sophie22:
ModuleNotFoundError: No module named 'distutils' - removed since Python 3.12, deprecated since Python 3.10

README.md
```diff
@@ -26,19 +26,19 @@ Quality assurance framework for Magnetic Resonance Imaging
...
 The hazen Tasks provide the following measurements within these phantoms:
 - Signal-to-noise ratio (SNR)
-- Spatial resolution
+- Spatial resolution (will not receive future updates)
 - Slice position
 - Slice width
 - Uniformity
 - Ghosting
-- MR relaxometry
+- MR relaxometry (will not receive future updates)
```
tomaroberts: I don't think we need to flag these. Prefer to to just keep the README clean.

requirements_Win3.11.txt
tomaroberts, line 1: What is the issue with the standard requirements.txt on Windows?

This one is enormous. There are a lot of unnecessary packages included, such as Jupyter and probably many more. Can this be shortened to just the necessary packages, like the standard one?

The PR says tested with Anaconda – have you tried it with pip on Windows?

tomaroberts (changes requested):
Couple of comments, mainly around the Windows requirements file.

requirements_Win3.11.txt
sophie22: Beatriz used pip on Windows and worked well for her. I tested with Anaconda because an issue was raised around that in particular, but turned out to be not version related.
Deleted this file for simplicity.

tomaroberts (approved):

PR #423, opened 2024-03-17
Ignore files in the input folder that have no pixel data

### LINKED ISSUE #408: Ignore files in the input folder that have no pixel data
**Describe the bug**
As the input to hazen is a directory path and directories may have various files, hazen errors out if it encounters a file that is not a DICOM or what only has DICOM header metadata, but no image (pixel array) data.
This is often the case when exporting QA images from Philips scanners which yields additional non-image DICOM files.

**Expected behavior**
Hazen should be able to determine and ignore files in the input folder that are not DICOM image files, and process suitable input files without errors or need to manually remove these files from the folder.

sophie22:
Added a new function to be part of the `get_dicom_files` function, which checks the folder items and only keeps the ones that have a parseable DICOM with a pixel_array. This is to avoid hazen breaking when subfolders or other files are present in the input folder.

These changes will pave the way for handling enhanced DICOM inputs in a future update.


hazenlib/tasks/ghosting.py
```diff
@@ -135,9 +135,6 @@ def get_signal_slice(self, bounding_box, slice_radius=5):
         )
         return idxs
 
-    def get_pe_direction(self, dcm):
```
sophie22, line 138: move function to utils.py

tests/test_enhanced.py
```diff
@@ -0,0 +1,86 @@
...
+
+import pydicom
+import unittest
+import hazenlib.utils as hazen_tools
+
+# class TestEnhanced(unittest.TestCase):
+def main():
+    TEST_DATA_DIR = "/Users/sophieratkai/Documents/projects/hazen_docs/ACR_se_enhanced"
+    enhanced_folder = os.path.join(TEST_DATA_DIR, 'enhanced')
+    folder = "/Users/sophieratkai/Documents/projects/hazen_docs/Philips_DicomTest/Res/Enhanced/DICOM"
```
tomaroberts: You have hard-coded filepaths from your local machine. Slightly puzzled why these haven't triggered as errors in the GitHub Actions. Can you check these tests are actually being run and passing, thanks.

tests/test_enhanced.py
```diff
@@ -0,0 +1,86 @@
...
+        # except:
+        #     print("enhanced", dcm.PerFrameFunctionalGroupsSequence[0].PixelMeasuresSequence[0].PixelSpacing)
+
+        # try:
+        #     print("RepetitionTime", dcm.RepetitionTime)
+        # except:
+        #     print("enhanced", dcm.SharedFunctionalGroupsSequence[0].MRTimingAndRelatedParametersSequence[0].RepetitionTime)
+
+
+    # def setUp(self):
```
tomaroberts: Are all these commented lines supposed to be here? Please update or remove.

tests/test_enhanced.py
```diff
@@ -0,0 +1,86 @@
...
+    
+
+
+if __name__ == '__main__':
+    main()
```
tomaroberts: Why is this here? test_*.py files shouldn't need such a statement.

tomaroberts (changes requested):
Mostly fine aside from test_enhanced.py – looks like an old file has been pushed or something possibly?

tests/test_enhanced.py
sophie22: copied from another test file we have in the folder.

PR #445, opened 2024-08-22
rotates sagittal acr images

rs-sprout98:
Rotates the slice thickness image 90 degrees clockwise if the image orientation is determined to be sagittal. Only works if original image is rotated 90 degrees anticlockwise.


mollybuckley (approved):
This is great Ryan! Approving these changes - thank you for doing this!

As discussed, we will temporarily update our local SOP with the correct direction to position the phantom - are you okay to follow that up with Becky?

As a next step, we can now look into detecting automatically whether the pixel array needs rotating clockwise or anti-clockwise and rotating accordingly.

PR #453, opened 2024-11-08
Update ACRObject.py with phantom rotation detection function.

xleyn:
Updated with a phantom rotation detection function.
The function takes a greyscale image as an input and outputs the phantom's rotation angle!
The code works in my local python software although this is my first time using github so apologies if this doesn't work without some tweaks. However, I thought it was worth submitting as a start!

### DIFF
--- a/hazenlib/ACRObject.py
+++ b/hazenlib/ACRObject.py
@@ -95,6 +95,31 @@ class ACRObject:
         # else:
         #     print("LR orientation swap not required.")
 
+    def calculate_rotation(img):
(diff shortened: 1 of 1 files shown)


mollybuckley (changes requested):
Hi Nathan, this is great - thank you very much for doing this!

One minor change- please could you add docstrings following the same format as those in the other functions?

Thank you!

xleyn:
Hi Molly, 

No problem! I've updated with the docstrings - anything else that needs changing?

Thanks, 
Nathan

PR #460, opened 2025-01-06
Slice thickness test xleyn

xleyn:
Resubmitting previous pull requests from a branch rather than from a local fork, as requested by Molly.

Main changes are:
1) Function added in ACRObject.py to accurately detect the angle of the phantom's rotation.
2) Functions in acr_slice_thickness.py to place two lines within the central insert of the phantom and calculate profiles across them. The lines are still placed accurately even when the phantom is rotated.


hazenlib/utils.py
abdrysdale, line 1: Just curious why the type annotations for Line and Point are string quoted and not using the object? That is, why this:

```python
def get_distance_to(self, other: "Point") -> float:
```

and not this:

```python
def get_distance_to(self, other: Point) -> float:
```

abdrysdale:
It also seems like a lot of redundant code could be removed by extending the numpy.ndarray base class?

Something like this for `Point` for example:

```python
import numpy as np
from typing import TypeVar

TPoint = TypeVar("TPoint", bound="Point")

class Point(np.ndarray):

    def __new__(cls, *args: float) -> np.ndarray:
        """Initialise the point class."""
        supported_dims = {2}
        if len(args) not in supported_dims:
            err = (
                f"Only {' or '.join(str(d) + 'D' for d in supported_dims)}"
                " points are supported"
            )
            raise NotImplementedError(err)
        return np.asarray(args, dtype=float).view(cls)

    def __init__(self, *args: float) -> None:
        """Initialise the point object."""
        self.x = args[0]
        self.y = args[1]
        self.xy = self

    def get_distance_to(self, other: TPoint) -> float:
        """Calculates distance between two point objects.

        Args:
            other : Point to calculate distance to.

        Returns:
            dist : The distance between the points.

        """
        if not isinstance(other, TPoint):
            err = f"other should be a Point and not {type(other)}"
            raise TypeError(err)
        return np.sqrt(np.sum((self - other)**2))

    def as_int(self) -> TPoint:
        """Returns an instance of Point with coords mapped to int.

        Returns:
            as_int : Instance of Point with coords mapped to int.

        """
        return self.astype(int)
```

Naturally the `xy` attribute isn't actually needed but just there for compatibility with your existing class.

I've also made `supported_dims` a set so that `Point` can be extended to 3D at some future point if need be.

*Updated with correct type annotations.*

hazenlib/utils.py
xleyn: Hi Alex, 

As far as I could tell, you can't directly type hint to the class that you're currently within. I believe that typing.Self was later added to solve this problem but as hazen requires 3.9, I changed the relevant annotations to strings. There may be a more appropriate solution though!

xleyn:
That's a great idea, thank you! Upon testing it out I've had some issues retaining attributes x and y after implicit additions of the point class (rather than custom __add__). I believe its because __init__ is not called in such a case. This can, of course, be solved by having x and y as properties rather than attributes but perhaps there is a better solution that I haven't been able to find as I am rather unfamiliar with using __new__

hazenlib/utils.py
abdrysdale: You're right! Thanks for explaining. Another solution could be to use `TypeVar`:

```python
from typing import TypeVar

T = TypeVar("T", bound="MyClass")

class MyClass:
    def method(self, other: T) -> list[T]:
        return (self, other)
```

which we might want to do if we ever subclass the `Point` class - I can't see a direct reason for this yet except maybe for the `Line` class. What do you think?

xleyn: I think that TypeVar does look like a cleaner solution. Personally, I can't see why Point would need to be subclassed but I reckon its better to implement TypeVar just in case as I wouldn't say it has too much of an impact on readability. 

Regarding the potential subclassing of Point for the Line class, I wonder whether its cleaner to just keep Line as a class that uses Point rather than inherits from it, as Line is not intuitively derived from Point. What do you think?

I have updated the code to reflect the changes that we've discussed - thanks for your help thus far!

rs-sprout98:
Hey Nathan - are you still available for Hazen? I've emailed you but the address I have keeps bouncing my emails back. If you can see this, send me a message on r.satnarine@nhs.net.

abdrysdale:
- Reimplemented in #504

Issue #463, opened 2025-01-16
Python 3.13 Compatibility 


abdrysdale:
- Run tests against Python Python 3.13.0

Issue #454, opened 2024-11-09
Include Python 3.12 in CI tests


abdrysdale:
Should be relatively simple to change the GitHub workflows to include 3.12

abdrysdale:
I'll have a look at #298 in more detail as there might be some overlap

Issue #448, opened 2024-10-01
TestACRSpatialResolutionGE::test_get_MTF50 fails on fresh install


abdrysdale:
On a fresh install of hazen, I get the following error:

```
$ pytest -k MTF50

================================ test session starts =================================
platform win32 -- Python 3.11.9, pytest-8.3.3, pluggy-1.5.0
rootdir: c:\Users\USER\Documents\programs\hazen
configfile: pyproject.toml
collected 205 items / 203 deselected / 2 selected

tests\test_acr_spatial_resolution.py .F                                         [100%]

====================================== FAILURES ======================================
_____________________ TestACRSpatialResolutionGE.test_get_MTF50 ______________________

self = <tests.test_acr_spatial_resolution.TestACRSpatialResolutionGE testMethod=test_get_MTF50>

    def test_get_MTF50(self):
        mtf50 = self.acr_spatial_resolution_task.get_mtf50(self.dcm)
        rounded_mtf50 = (np.round(mtf50[0], 2), np.round(mtf50[1], 2))
    
        print("\ntest_get_MTF50.py::TestGetMTF50::test_get_MTF50")
        print("new_release_value:", rounded_mtf50)
        print("fixed_value:", self.MTF50)
    
>       assert rounded_mtf50 == self.MTF50
E       assert (0.5, 0.91) == (0.5, 1.85)
E         
E         At index 1 diff: 0.91 != 1.85
E         Use -v to get more diff

tests\test_acr_spatial_resolution.py:70: AssertionError
-------------------------------- Captured stdout call --------------------------------

test_get_MTF50.py::TestGetMTF50::test_get_MTF50
new_release_value: (0.5, 0.91)
fixed_value: (0.5, 1.85)
================================== warnings summary ==================================
tests/test_acr_spatial_resolution.py::TestACRSpatialResolutionGE::test_get_MTF50
  c:\Users\USER\Documents\programs\hazen\hazenlib\tasks\acr_spatial_resolution.py:261: OptimizeWarning: Covariance of the parameters could not be estimated
    popt, pcov = scipy.optimize.curve_fit(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================== short test summary info ===============================
FAILED tests/test_acr_spatial_resolution.py::TestACRSpatialResolutionGE::test_get_MTF50
=============== 1 failed, 1 passed, 203 deselected, 1 warning in 7.58s ===============
```

Are you able to reproduce this error?

[...]

PR #479, opened 2025-07-31
Updated to Python 3.13 (with migrating to uv and fixing failing tests)

### LINKED ISSUE #477: Migrate from setup.py to uv
[uv](https://docs.astral.sh/uv/) is replacement for pip, pip-tools, pipx, poetry, pyenv, twine, virtualenv, and more. It is quickly becoming the defacto package and project manager for Python with similar functionality to Cargo's rust.

Moreover, [PEP 737](https://peps.python.org/pep-0735/) advises against using `requirements.txt` and [PEP 518](https://peps.python.org/pep-0518/) recommends using the build system in `pyproject.toml`.

As far as I can tell migrating to `uv` would solve the following currently open issues:

- #463 
- #454 
- #430 

It would mean run tests on a fresh install would look like this:

```sh
git clone git@github.com:GSTT-CSC/hazen.git
uv run pytest
```

and `uv` would create a virtual environment, install the correct version of python, install the required dependencies and run the tests.

Also, unlike the current build system, `uv` is truly cross platform, working equally well for UNIX and Windows systems.

I'm happy to do this as I've just finished migrating all of our teams project to `uv`.

What are your thoughts?

abdrysdale:
- Closes #477 #463 #454 #448

- Migrated to uv along with including Python versions 3.9 to 3.13 (inclusive) in build and tests.
- With this, packages were migrated to newest versions which included replacing legacy functions (such as `pydicom.pixel_data_handelrs.util` in favour of `pydicom.pixels` and `np.int0` with `np.intp`).
- Similarly, as numpy C API has changed the tests have been changed to use `np.testing.assert_allclose` which is more robust to subtle errors due to floating point arithmetic.
- Workflows have been updated to use uv instead of legacy setuptools.
- The release package name is now set based of the repo name (e.g. hazen for GSTT-CSC/hazen and hazen-wales for sbu-physics-mri/hazen-wales) to allow us to keep an update-to-date version of our own package on PyPI whilst ensuring that all changes can be contributed back to this central repository.


abdrysdale:
Closed pending PR from @luissantosHCIT

PR #495, opened 2025-12-02
Rotation function.

DanielJohnWest:
Rotation function to ensure correct phantom orientation across tests (SNR and uniformity fully working), PR 460 used for slice thickness, but slice position and geometric accuracy require further refinement (relating to other open issues).

### DIFF
--- a/hazenlib/tasks/acr_geometric_accuracy.py
+++ b/hazenlib/tasks/acr_geometric_accuracy.py
@@ -26,6 +26,7 @@ import os
 import sys
 import traceback
 import numpy as np
+import pydicom
 
 import skimage.measure
 import skimage.transform
@@ -33,7 +34,9 @@ import skimage.morphology
 
 from hazenlib.HazenTask import HazenTask
 from hazenlib.ACRObject import ACRObject
+from hazenlib.logger import logger
 
+from scipy.ndimage import center_of_mass, shift as nd_shift
 
 class ACRGeometricAccuracy(HazenTask):
     """Geometric accuracy measurement class for DICOM images of the ACR phantom."""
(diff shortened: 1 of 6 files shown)


abdrysdale:
Thanks for the PR!

I've just had a quick look (looks great) and will properly review later. It looks like the unittests fail - do they fail your end? I imagine this is expected but it would be good to have some justification for changing test values in the form of a commit.

hazenlib/tasks/acr_geometric_accuracy.py
```diff
@@ -42,6 +45,65 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.ACR_obj = ACRObject(self.dcm_list)
 
+    def detect_best_rotation(self, ref_img):
```
abdrysdale: Rather than adding `detect_best_rotation` as a method for every class I think it would best to move this to a method for the ACR object.

This way we only have one implementation and if we ever need to make changes to this method in the future we only need to change one method rather than one method for every class - I hope that makes sense.

hazenlib/tasks/acr_geometric_accuracy.py
```diff
@@ -42,6 +45,65 @@ def __init__(self, **kwargs):
...
+
+            if mse < best_mse:
+                best_mse = mse
+                best_name = name
+                best_func = func
+
+        logger.info(f"Best rotation/flip detected: {best_name} (MSE={best_mse:.4e})")
+        return best_name, best_func
+
+    def apply_rotation_to_stack(self, rotation_func):
```
abdrysdale: Similarly, it would be best if this was defined in one place rather than for every class. It also looks like there is some overlap with the method `rotate_images` in the `ACRObject.py` file.

hazenlib/tasks/acr_uniformity.py
```diff
@@ -77,16 +144,16 @@ def get_integral_uniformity(self, dcm):
         """
         img = dcm.pixel_array
         # Required pixel radius to produce ~200cm2 ROI
-        r_large = np.ceil(80 / self.ACR_obj.dx).astype(int)
```
abdrysdale: Just curious for the change in this value (from 80 -> 75)? Should this be a keyword argument?

hazenlib/utils.py
```diff
@@ -460,7 +455,9 @@ def determine_orientation(dcm_list):
             return "axial", z
         else:
             logger.warning("Unable to determine orientation based on DICOM metadata")
-            logger.info("x %s\ny %s\nz %s", set(x), set(y), set(z))
```
abdrysdale: I think the original implementation here is more concise so I would revert back to that. Is there a reason for splitting up the logging messages?

PR #504, opened 2026-03-17
Slice thickness v2

abdrysdale:
- New implementation of #460 originally by @xleyn for the version 2 codebase.


abdrysdale:
@rs-sprout98 or @DanielJohnWest would you mind reviewing this PR? It should be fully merge-able and the methodology makes sense to me. The test values are quite different to Louis' method so I just wanted a second opinion before updating the test values.

rs-sprout98 (dismissed):

